PR #33, opened 2019-01-08
Auto-generate multus conf. Do not exit sriov DS

krsna1729:
Instead of relying on configmap to be created/customized by the user to
generate multus conf, we autogenerate it using the current default CNI
conf file on the host. This way CNI's with host specific conf details
will continue to work.

```
{
  "type": "calico",
  "nodename": "clr-01",
   ...
}
```

Since SR-IOV device may not be present on all nodes, the daemonset
should not exit as this will put the pod in restart loop. This would
cause the kubelet and runtimes to repeatedly restart, as per
initContainer logic.


clr-k8s-examples/9-multi-network/multus-sriov-ds.yaml
```diff
@@ -87,6 +87,20 @@ data:
         user: multus
     current-context: multus-context
     EOF
+  install-multus-conf.sh: |
```
mcastelino: @krsna1729 is this cleaned up? What will happen the next time this script runs? Will it not see 00-multus.conf at /host/etc/cni/net.d

krsna1729: @mcastelino  Ooh thats right. Good catch.

krsna1729:
@mcastelino addressed.

mcastelino:
LGTM

PR #36, opened 2019-01-17
Change mount dir inside vagrant

krsna1729:


krsna1729:
Primarily for people not too well versed in vagrant and where to look. Immediately one would do `ls` in the folder hoping to find something. Tested with the change, no issues.

PR #41, opened 2019-01-29
Switch vagrant boxes

ganeshmaharaj:
Switching boxes to the one maintained by Antonio at
https://github.com/AntonioMeireles/ClearLinux-packer. They now support
libvirt and will eventually be part of the clearlinux tree.


krsna1729:
@ganeshmaharaj can we remove this
```
      # Bad hack for the vagrant libvirt boxes. WIll be removed once they are fixed.
      c.vm.provision "shell", privileged: false, inline: "sudo usermod --password vagrant root"
```

AntonioMeireles:
> @ganeshmaharaj can we remove this
> 
> ```
>       # Bad hack for the vagrant libvirt boxes. WIll be removed once they are fixed.
>       c.vm.provision "shell", privileged: false, inline: "sudo usermod --password vagrant root"
> ```

for my boxes yes... default root passwd  is `V@grant!`

PR #45, opened 2019-01-31
Change disk image names

ganeshmaharaj:
Avoid creating unique names for disks as vagrant-libvirt today does not
delete those images when VMs are deleted. Once that is fixed we can move
back to using unique names. Until then this should suffice.


krsna1729:
LGTM

PR #48, opened 2019-02-01
Add containerd

egernst:
Add support for running create_stack and reset_stack for containerd.

Initial script for installing and configuring containerd.

An ideal next step would be to update service file for kubelet to take the new socket path into account, but this is currently left up to user (since it varies slightly pending the distro ... PRs welcome!)


clr-k8s-examples/create_stack.sh
```diff
@@ -112,20 +114,37 @@ function all() {
...
-	print_usage_exit
-	;;
-esac
+	cd $SCRIPT_DIR
+
+
+	while getopts r: opt; do
+		case $opt in
+		r)
+			runtime="$OPTARG"
```
mythi: would be good to validate the input strings

clr-k8s-examples/containerd-setup/configure-containerd.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+VERISION="1.2.2"
+echo "Install Containerd ${VERSION}"
+wget -q https://storage.googleapis.com/cri-containerd-release/cri-containerd-${VERSION}.linux-amd64.tar.gz
```
mythi: Let's not do this but update the version in Clear. I'm not aware of blockers for that. Do you know any?

egernst: Can I *just* install containerd?

mythi: Clear already has two versions of it available. Neither are good for runtimeclass but my feedback was that we should do the update there so that pulling the third version is not needed.

krsna1729:
@egernst Need doc update too

clr-k8s-examples/8-kata/runc-runtimeClass.yaml
```diff
@@ -0,0 +1,6 @@
+kind: RuntimeClass
```
mcastelino: @egernst why was this needed. Does containerd require runc to be explicitly listed?

clr-k8s-examples/containerd-setup/configure-containerd.sh
mcastelino: @egernst I agree with @mythi . Let us just ensure that the containerd we ship with clearlinux works with runtime class. That will ensure that it is updated as part of swupd. Otherwise it has to be maintained and updated independently.

/cc @bryteise

clr-k8s-examples/containerd-setup/containerd.conf
```diff
@@ -0,0 +1,21 @@
...
+    no_prometheus = false
+    [plugins.cri.containerd]
+      snapshotter = "overlayfs"
+      no_pivot = false
+      [plugins.cri.containerd.runtimes.runc]
+        runtime_type = "io.containerd.runtime.v1.linux"
+        [plugins.cri.containerd.runtimes.runc.options]
+          Runtime = "runc"
+          RuntimeRoot = "/usr/local/sbin/runc"
+      [plugins.cri.containerd.runtimes.kata-qemu]
```
mcastelino: @egernst should we just use the shim-v2 version. It will allow us to skip jumping through the bash scripts as container-shim-v2 already supports passing args to the runtime binary I think.

ganeshmaharaj:
@egernst @mcastelino we good on this one? should we get this merged?

clr-k8s-examples/containerd-setup/configure-containerd.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+VERISION="1.2.2"
```
ganeshmaharaj: Typo

clr-k8s-examples/containerd-setup/configure-containerd.sh
```diff
@@ -0,0 +1,17 @@
...
+echo "Install Containerd ${VERSION}"
+wget -q https://storage.googleapis.com/cri-containerd-release/cri-containerd-${VERSION}.linux-amd64.tar.gz
+sudo tar -C / -xzf cri-containerd-${VERSION}.linux-amd64.tar.gz
+
+# configure runtime classes for Kata
+mkdir -p /etc/containerd
```
ganeshmaharaj: This also should be run with sudo.

mcastelino:
> @egernst @mcastelino we good on this one? should we get this merged?

Once you are able to test that it works we can merge it. But longer term we should just package the right version of containerd as part of cloud-native-basic and containers-basic bundles.

Also we need to run the kubernetes conformance tests against this setup.

But both those things can be done in follow up PRs.

ganeshmaharaj:
@egernst it might also be good if you can add containerd in the proxy configuration setup in `setup_system.sh` script.

ganeshmaharaj:
This stuff is now in and we have containerd support. Think we can go ahead and close this PR. @egernst @krsna1729 @mcastelino ?

Issue #38, opened 2019-01-27
Use kata-deploy to install all kata components


krsna1729:
Remove [`setup_kata_firecracker.sh`](https://github.com/clearlinux/cloud-native-setup/blob/master/clr-k8s-examples/setup_kata_firecracker.sh) once https://github.com/kata-containers/packaging/pull/304 merges and use kata-deploy going forward

krsna1729:
@egernst @mcastelino @grahamwhaley

krsna1729:
@egernst should #39 be included in kata-deploy

PR #50, opened 2019-02-05
Use kata-deploy to install and configure kata

krsna1729:
Resolves: #38 

TODO:
Fix FC failing

```
NAME                                   READY   STATUS              RESTARTS   AGE
php-apache-kata-fc-6c6f484c4b-s5rz7    0/1     ContainerCreating   0          4m45s
php-apache-kata-qemu-bc4c55994-4vh4b   1/1     Running             0          4m45s
php-apache-runc-7cc9dcb54f-4sv2c       1/1     Running             0          4m45s
test-cpumanager-kata                   0/1     Completed           0          4m45s
test-cpumanager-runc                   0/1     Completed           0          4m45s
```
```
  Warning  FailedCreatePodSandBox  8s (x5 over 82s)  kubelet, clr-01    Failed create pod sandbox: rpc error: code = Unknown desc = container create failed: Failed to check if grpc server is working: context deadline exceeded
```

### DIFF
--- a/clr-k8s-examples/8-kata/deploy/kata-cleanup.yaml
+++ b/clr-k8s-examples/8-kata/deploy/kata-cleanup.yaml
@@ -1,0 +1,46 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kubelet-kata-cleanup
+  namespace: kube-system
+spec:
+  selector:
+      matchLabels:
+        name: kubelet-kata-cleanup
+  template:
+    metadata:
+        labels:
+          name: kubelet-kata-cleanup
+    spec:
+      serviceAccountName: kata-label-node
+      nodeSelector:
+          katacontainers.io/kata-runtime: cleanup
+      containers:
+      - name: kube-kata-cleanup
+        image: katadocker/kata-deploy
+        imagePullPolicy: Always
+        command: ["/bin/bash", "-c", "/opt/kata-artifacts/scripts/kata-deploy.sh reset"]
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        securityContext:
+          privileged: false
+        volumeMounts:
+        - name: dbus
+          mountPath: /var/run/dbus
+        - name: systemd
+          mountPath: /run/systemd
+      volumes:
+        - name: dbus
+          hostPath:
+            path: /var/run/dbus
+        - name: systemd
+          hostPath:
+            path: /run/systemd
+  updateStrategy:
+    rollingUpdate:
+      maxUnavailable: 1
+    type: RollingUpdate
--- a/clr-k8s-examples/8-kata/deploy/kata-deploy.yaml
+++ b/clr-k8s-examples/8-kata/deploy/kata-deploy.yaml
@@ -1,0 +1,64 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kata-deploy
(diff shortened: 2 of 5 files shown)


krsna1729:
The issue is with vsock and not this patch. qemu also fails when told to `use_vsock`. removing DNM label.

@mcastelino PTAL and merge

ganeshmaharaj:
@mcastelino, are there any clear specific stuff for kata-fc that might hinder things working if we use the ones shipped with kata-deploy?

mcastelino:
> @mcastelino, are there any clear specific stuff for kata-fc that might hinder things working if we use the ones shipped with kata-deploy?

The benefit of using the firecracker that comes with clear is that it will updated as clear updates with any security and other fixes. The same reason for using containerd packaged with clear.

But then we will split across deployment methodologies.

@egernst @krsna1729 what is our update model for kata-deploy?

egernst:
> what is our update model for kata-deploy?
@mcastelino - we update the binaries on a per release basis (scheduled every 2 weeks, though quicker if issues are identified)

krsna1729:
@mcastelino @ganeshmaharaj @egernst @grahamwhaley @bryteise 
Time to think about how the infra is supposed to be upgraded from node and this stack perspective

mythi:
@krsna1729 kubeadm has documentation how to use ```kubeadm upgrade``` to update the cluster and we have a tutorial how to take that in Clear/swupd world.

Back in when we changed 1.12->1.13 I used the tutorial to test it's possible to update a cluster running this stack. I did not look kata specific things then tough.

krsna1729:
Created #53 to capture

ganeshmaharaj:
Tested this patch and +1 after two things happen.

https://github.com/clearlinux/cloud-native-setup/pull/59 <-- that fixes the issue where kata-runtime definition in crio.conf by clear does not get overwritten.
https://github.com/kata-containers/packaging/pull/334 <-- So that network_ns_lifecycle string is not duplicated and breaks crio service.

krsna1729:
@ganeshmaharaj we should move this block to `setup_system.sh` in this patch if we want to use kata-deploy right?

```
sudo mkdir -p /etc/crio/
if [ ! -f /etc/crio/crio.conf ]; then
  sudo cp /usr/share/defaults/crio/crio.conf /etc/crio/crio.conf
fi
```

ganeshmaharaj:
@krsna1729 we can do the move as part of a clean-up patch where we will also flush the vagrant file. for now, this should be alright to have. if there are no objections, we can merge this.

krsna1729:
@mcastelino?
cc @AntonioMeireles for insights

AntonioMeireles:
@sai

 fwiw i find somewhat ironic the use of kata-deploy to install kata given
that imho the current cloud-native-setup structural design flaw is the
insistence to use the in-distro kubelet/kubeadm/kubectl which basically
just breaks updating at will . in other words kata-deploy is much needed
progress in right direction until swupd featureset improved in this very
specific regard 😃

mythi:
> insistence to use the in-distro kubelet/kubeadm/kubectl which basically just breaks updating at will

I've successfully updated a cluster with the swupd model when the k8s minor version changes (i.e., API compatibility breaks) and it works. cloud-native-setup does not mandate to use in-distro k8s binaries as far as I can tell (you just need to have them installed). What problems do you see with updates today (other than the lack of version pinning which isn't cloud-native bundle specific)?

krsna1729:
needs rebase

krsna1729:
rebased and verified working with upstream kata-deploy latest image

Issue #5, opened 2018-12-04
HPA: Metrics server may not be working


krsna1729:
Testing as per [`test-autoscale.sh`](cloud-native-setup/clr-k8s-examples/tests/test-autoscale.sh). The pods are not scaling.
```
Name:                                                  php-apache-test
Namespace:                                             default
Labels:                                                <none>
Annotations:                                           <none>
CreationTimestamp:                                     Tue, 04 Dec 2018 22:36:01 +0000
Reference:                                             Deployment/php-apache-test
Metrics:                                               ( current / target )
  resource cpu on pods  (as a percentage of request):  <unknown> / 50%
Min replicas:                                          1
Max replicas:                                          10
Deployment pods:                                       1 current / 0 desired
Conditions:
  Type           Status  Reason                   Message
  ----           ------  ------                   -------
  AbleToScale    True    SucceededGetScale        the HPA controller was able to get the target's current scale
  ScalingActive  False   FailedGetResourceMetric  the HPA was unable to compute the replica count: unable to get metrics for resource cpu: no metrics returned from resource metrics API
Events:
  Type     Reason                        Age                From                       Message
  ----     ------                        ----               ----                       -------
  Warning  FailedComputeMetricsReplicas  27s (x12 over 3m)  horizontal-pod-autoscaler  failed to get cpu utilization: unable to get metrics for resource cpu: no metrics returned from resource
metrics API
  Warning  FailedGetResourceMetric       12s (x13 over 3m)  horizontal-pod-autoscaler  unable to get metrics for resource cpu: no metrics returned from resource metrics API
```

PR #54, opened 2019-02-08
Update to latest metrics-server 0.3.x works

krsna1729:
With kubelet service file fixed and correct args set in
metrics-server-deployment.yaml, autoscale seems to be working reliably.

Fixes: #5


krsna1729:
```
$ cat /etc/os-release
NAME="Clear Linux OS"
VERSION=1
ID=clear-linux-os
ID_LIKE=clear-linux-os
VERSION_ID=27700
PRETTY_NAME="Clear Linux OS"
```

krsna1729:
```
./create_stack.sh minimal
./tests/autoscale/test-autoscale.sh
```

clr-k8s-examples/tests/autoscale/load-gen.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Pod
```
mcastelino: @krsna1729 should we make this a deployment. That way we can scale it up to generate more load?

clr-k8s-examples/1-core-metrics/metrics-server-deployment.yaml
```diff
@@ -22,10 +23,19 @@ spec:
...
       - name: metrics-server
-        image: gcr.io/google_containers/metrics-server-amd64:v0.2.0
+        image: k8s.gcr.io/metrics-server-amd64:v0.3.1
         imagePullPolicy: Always
-        command:
-        - /metrics-server
-        - --source=kubernetes.summary_api:https://kubernetes.default.svc?kubeletHttps=true&kubeletPort=10250&useServiceAccount=true&insecure=true
+        args:
+        - --logtostderr
+        - --kubelet-insecure-tls
```
mcastelino: @krsna1729 is this a limitation. This was due to a bug right.

mcastelino:
@ganeshmaharaj @krsna1729 can we spin a new release of this package on Clearlinux. This gives us the the metrics server we wanted. We had downgraded the metrics server due to issues we had seen.

clr-k8s-examples/1-core-metrics/metrics-server-deployment.yaml
krsna1729: This was because of self-signed certs IIRC

PR #57, opened 2019-02-13
Downloading OVMF causes issues with libvirt

mcastelino:
The parent box is already downloading the OVMF firmware.
Just use the firmware that is already downloaded.


krsna1729:
@mcastelino @ganeshmaharaj so where is loader being used?

PR #59, opened 2019-02-17
Check crio.conf before copying

ganeshmaharaj:

### DIFF
--- a/clr-k8s-examples/setup_kata_firecracker.sh
+++ b/clr-k8s-examples/setup_kata_firecracker.sh
@@ -74,7 +74,9 @@ sudo chmod +x /usr/bin/kata-fc
 # Also setup crio to use devicemapper
 
 sudo mkdir -p /etc/crio/
-sudo cp /usr/share/defaults/crio/crio.conf /etc/crio/crio.conf
+if [ ! -f /etc/crio/crio.conf ]; then
+  sudo cp /usr/share/defaults/crio/crio.conf /etc/crio/crio.conf
+fi
 
 echo -e "\n[crio.runtime.runtimes.kata-qemu]\nruntime_path = \"/usr/bin/kata-runtime\"" | sudo tee -a /etc/crio/crio.conf
 echo -e "\n[crio.runtime.runtimes.kata-fc]\nruntime_path = \"/usr/bin/kata-fc\"" | sudo tee -a /etc/crio/crio.conf


krsna1729:
LGTM

mythi:
Joining late. 

Can we get rid of ```crio.conf```` copying? It used to be so that kata was separate from cloud-native-basic bundle but it's now installed part of it. 

With that, our default ```crio.conf``` could have all runtimes configured by default. After all the selection comes from ```RuntimeClass```

mythi:
````kata-runtime``` package copies it too. What are 'all the changes' you're referring to?

**Comment by ganeshmaharaj** [comment]
this patch, there is one to fix kata-deploy and there is a pending clean-up patch for all the setups we are doing with cloud-native-setup. are there other places where we are doing the copying? if so, can you share that list here and we can work through them to get this all fixed up.

**Comment by mythi** [comment]
here: https://github.com/clearlinux-pkgs/kata-runtime/blob/master/0002-Set-kata-runtime-as-default-runtime-for-cri-o.patch

**Comment by krsna1729** [comment]
@mythi what am I missing?

```
clear@clr-01 ~ $ sudo swupd bundle-add cloud-native-basic
Downloading packs...

Extracting cloud-native-basic pack for version 27870
        ...100%
Starting download of remaining update content. This may take a while...
        ...100%
Finishing download of update content...
Installing bundle(s) files...
        ...100%
Calling post-update helper scripts.
Successfully installed 1 bundle
clear@clr-01 ~ $ ls /etc/crio/crio.conf
ls: cannot access '/etc/crio/crio.conf': No such file or directory
```

**Comment by mythi** [comment]
@krsna1729 it gets copied by ```crio-set-runtime.service``` if all conditions are met.

**Comment by krsna1729** [comment]
More guidance/detail would help
```
clear@clr-01 ~ $ ls /dev/kvm
/dev/kvm
clear@clr-01 ~ $ systemctl status crio-set-runtime.service
● crio-set-runtime.service - Set Default Cri-o Runtime
   Loaded: loaded (/usr/lib/systemd/system/crio-set-runtime.service; static; vendor preset: disabled)
   Active: inactive (dead)
clear@clr-01 ~ $ sudo systemctl enable --now crio-set-runtime.service
The unit files have no installation config (WantedBy, RequiredBy, Also, Alias
settings in the [Install] section, and DefaultInstance for template units).
This means they are not meant to be enabled using systemctl.
Possible reasons for having this kind of units are:
1) A unit may be statically enabled by being symlinked from another unit's
   .wants/ or .requires/ directory.
2) A unit's purpose may be to act as a helper for some other unit which has
   a requirement dependency on it.
3) A unit may be started when needed via activation (socket, path, timer,
   D-Bus, udev, scripted systemctl call, ...).
4) In case of template units, the unit is meant to be enabled with some
   instance name specified.
clear@clr-01 ~ $
```

mythi:
@krsna1729 Perhaps it's broken :) crio.service ```Wants``` it so simply restarting crio should do it.

PR #60, opened 2019-02-20
Fix for duplicate lifecycle entry in crio.conf

ganeshmaharaj:


ganeshmaharaj:
@krsna1729 while we wait on kata-deploy to be ready, the previous patch has broken the current setup. Thinking we can land this and then rebase kata-deploy when the new container images are ready.

PR #66, opened 2019-03-12
Disable proxy

mcastelino:
Disable proxy setup if the host does not have proxy setup in
the enviornment variables


AntonioMeireles:
(hi!)
to play extra safe why not ?...
```ruby
if ENV['http_proxy'] || ENV['HTTP_PROXY']
  if Vagrant.has_plugin?("vagrant-proxyconf")
   (...) 
```

PR #68, opened 2019-03-13
 Update webhook setup and documentation

mcastelino:
The admission controller is now able to access full context. Hence explicit tagging of pod manifests to exclude them from kata is no longer required.

Additonally, we only support kubernetes versions with support for runtimeClass.  Remove the legacy annotations to reduce confusion.

### DIFF
--- a/clr-k8s-examples/7-rook/000-operator.yaml
+++ b/clr-k8s-examples/7-rook/000-operator.yaml
@@ -2,8 +2,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: rook-ceph-system
-  labels:
-    kata: "false"
 ---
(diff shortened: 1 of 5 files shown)


krsna1729:
@mcastelino can we also pull in the `deploy` folder from upstream and the script?

mcastelino:
> @mcastelino can we also pull in the deploy folder from upstream and the script?

@krsna1729 unfortunately we need the generated manifests. The https://github.com/kata-containers/tests/tree/master/kata-webhook/deploy does not have the generated manifests.

krsna1729:
@mcastelino instead of just apply, we will call the script that generates and apply. We will be pulling both script and deploy folder

mcastelino:
@krsna1729 I will do that in a follow up. I need to change the kata tests.

PR #74, opened 2019-03-20
Do not restart containerd, remove stop kubelet

krsna1729:
In case of containerd there is no need to restart. Restart for crio is needed
when it manages the networks ns lifecycle. Removed the kubelet stop step which
was a leftover from earlier device plugin install.


mcastelino:
@krsna1729 is this a bug in containerd. You should be able to restart containerd.

PR #76, opened 2019-03-26
Update README.md

krsna1729:
Add note about environment variables that can be used to customize the setup


grahamwhaley:
Thanks @krsna1729 - I thought we might gain a section noting that if you are running a single node cluster (that is, just using the master node and not joining the slave nodes), then you *will* have to increase the defaults I think to have it stable?
But, as we don't actually have a single node cluster section at present, np.

PR #78, opened 2019-03-27
Turn on static policy for CPU manager

### LINKED ISSUE #71: Should we enable --cpu-manager-policy=static by default
The static policy allows containers in Guaranteed pods with integer CPU requests access to exclusive CPUs on the node. This exclusivity is enforced using the cpuset cgroup controller.

https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/

krsna1729:
This is finally supported in Kata 1.6
Update memory limit for kata to run in guaranteed test case

Resolves #71 

### DIFF
--- a/clr-k8s-examples/kubeadm.yaml
+++ b/clr-k8s-examples/kubeadm.yaml
@@ -7,9 +7,8 @@ apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
 featureGates:
   RuntimeClass: true
-# Kata does not work with static
-# https://github.com/kata-containers/runtime/issues/878
-# cpuManagerPolicy: static
+# Allowing for CPU pinning and isolation in case of guaranteed QoS class
+cpuManagerPolicy: static
 systemReserved:
   cpu: 500m
   memory: 256M
--- a/clr-k8s-examples/tests/cpumanager/test-cpumanager.yaml.tmpl
+++ b/clr-k8s-examples/tests/cpumanager/test-cpumanager.yaml.tmpl
@@ -12,7 +12,7 @@ spec:
     resources:
       limits:
         cpu: 1
-        memory: 100Mi
+        memory: 500Mi # For kata to run
 ---
 apiVersion: v1
 kind: Pod


clr-k8s-examples/kubeadm.yaml
```diff
@@ -7,9 +7,7 @@ apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
 featureGates:
   RuntimeClass: true
-# Kata does not work with static
-# https://github.com/kata-containers/runtime/issues/878
-# cpuManagerPolicy: static
+cpuManagerPolicy: static
```
grahamwhaley: Great that we now support this. A couple of questions:
 - Do I remember you noting that there is still maybe one aspect or feature that is not fully supported? Or, was that fixed by the other addition here (the memory increase) ?
- afaict, the [k8s default](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#configuration) is `none`. If we are changing the default here for an explicit reason, can we add back the comment and explain why?

krsna1729: Kata required more memory so test case issue. Fixed.

About Cpumanagerpolicy static, we can say -

To allow for CPU pinning and isolation of guaranteed workloads

jcvenegas: @grahamwhaley the memory update is related with this error:
https://github.com/kata-containers/runtime/issues/1249

qemu is running the pod cgroup, this is too small to run kata, and qemu is killed at boot.

Related static policy this is a feature we want to make sure it works with kata (or is the most compatible as possible) . Today is not fully supported as is expected from runc perspective but at least the cpuset is honor at host so isolation is guaranteed at node level. There are some missing gaps from workload and performance perspective but having enabled helps to not break the use case on the future again.

mcastelino: @jcvenegas CPU Set is not fully honored right? On the host at the VM level. However I would still like to merge this as it allows runc to work properly. 

We need to ensure the Kata does not trample on runc pods. The assumption will be that we will not use cpu sets for kata (CFS only). That is the most import test we need to pass. Kata still managed via CFS and cpu sets on a case by case basis for runc?

@krsna1729 agree?

mcastelino: /cc @mythi

PR #81, opened 2019-03-28
kata-deploy: update to match what's on Kata master

egernst:
We updated to use containerd v2 shim - pull in latest from master

Noticed there's been a discrepancy in the yaml between Kata and this repo for /bin/bash v. bash.  PTAL at this @krsna1729 ?


krsna1729:
@egernst should not make a difference.

Issue #84, opened 2019-04-04
error running create_stack.sh


kiendinh:
I'm using the master branch @f863bb7f, and running create_stack.sh returned with this error:

`daemonset.apps/kata-deploy created
error: the path "8-kata/runtimeclass_crd.yaml" does not exist`

Looks like the latest commit removes the file that is required elsewhere. Could anyone please check?

krsna1729:
@kiendinh Thanks for catching that. My mistake. Will test and fix up.

PR #85, opened 2019-04-04
Remove applying runtimeclass CRD

krsna1729:
CRD was removed as part of 1.14 update. Missed update to create_stack

Fixes: #84

### DIFF
--- a/clr-k8s-examples/create_stack.sh
+++ b/clr-k8s-examples/create_stack.sh
@@ -42,13 +42,6 @@ function kata() {
 	# Install kata artifacts using kata-deploy
 	kubectl apply -f 8-kata/deploy/kata-rbac.yaml
 	kubectl apply -f 8-kata/deploy/kata-deploy.yaml
-
-	#Add support for kata runtime
-	kubectl apply -f 8-kata/runtimeclass_crd.yaml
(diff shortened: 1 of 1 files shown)


mcastelino:
@krsna1729 @ganeshmaharaj we should update our readme to say our scripts will only work on 1.14 and beyond.

So as long as the user uses the files from the bundle vs here they should be fine.

@ganeshmaharaj did you all ensure the scripts got updated in the clearlinux cloud-native-basic example scripts. We package the scripts today as part of CL

/cc @bryteise

krsna1729:
@mcastelino @bryteise  yes we need this updated in the bundle

ganeshmaharaj:
+1. Manohar we should cut a release now and i can update the scripts in clear. Anything we would like to add before i cut the release?

mcastelino:
@ganeshmaharaj @bryteise yes we need to cut a new release of the cloud-native-basic bundle with the updated scripts. Otherwise the scripts will not work with the version of kubernetes tools and kubelet packaged. We need to do this each time we rev any of the cloud-native-basic components.

ganeshmaharaj:
fair enough. We don't have a auto-trigger of sorts to do this today. Needs a human tracking it. We need to figure out how to automate this.

PR #86, opened 2019-04-04
Fix readme to add k8s version dependency

ganeshmaharaj:

### DIFF
--- a/clr-k8s-examples/README.md
+++ b/clr-k8s-examples/README.md
@@ -1,5 +1,8 @@
 # How to setup the cluster
 
+## Prerequisite
+This setup currently will work with k8s 1.14 & above. Any version of k8s before that might work, but is not guaranteed.
+
 ## Sample multi-node vagrant setup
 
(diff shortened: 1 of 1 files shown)


krsna1729:
@mcastelino i dont understand why we need this. We said there is no pinning. we continue to work on whatever is packaged on latest clear. and if anything breaks we fix.

mcastelino:
@krsna1729 these scripts will work with the latest version of k8s. However Clearlinux itself carries a copy of these scripts. Those scripts should work with the version of kubernetes that Clearlinux carries.

So when Clearlinux updates any of the cloud-native-basic components the scripts packaged within that package in Clearlinux should work. So if an user were to use the scripts packaged with Clear, they should work.

That allows this rep to move and track latest, while an user can move the Clearlinux version forward or backward and always have working scripts.

Does that make sense?

ganeshmaharaj:
@mcastelino @krsna1729 point is fair. i agree we need to package the right version of the scripts to  clear but do not need to put that in the notes. We can make a tag of the package and just push an update without this change. This disclaimer will be true if we folks use these scripts in other OSes apart from clear, at which point it won't hurt us to have this.

PR #95, opened 2019-05-24
stack: dont provide CRI socket

jcvenegas:
kubeadm autodetects the socket path based on defaults from well known
CRI servers.

Signed-off-by: Jose Carlos Venegas Munoz <jose.carlos.venegas.munoz@intel.com>

### DIFF
--- a/clr-k8s-examples/kubeadm.yaml
+++ b/clr-k8s-examples/kubeadm.yaml
@@ -1,7 +1,5 @@
 apiVersion: kubeadm.k8s.io/v1beta1
 kind: InitConfiguration
-nodeRegistration:
-  criSocket: /var/run/crio/crio.sock
 ---
 apiVersion: kubelet.config.k8s.io/v1beta1
 kind: KubeletConfiguration
--- a/clr-k8s-examples/reset_stack.sh
+++ b/clr-k8s-examples/reset_stack.sh
@@ -3,7 +3,10 @@
 set -o nounset
 
 #Cleanup
-sudo -E kubeadm reset -f --cri-socket="/var/run/crio/crio.sock"
+reset_cluster() {
+	sudo -E kubeadm reset -f
+}
+reset_cluster
 
 for ctr in $(sudo crictl ps --quiet); do
 	sudo crictl stop "$ctr"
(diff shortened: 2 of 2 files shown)


ganeshmaharaj:
@jcvenegas might i ask you to take a quick peek at the containerd patch and see if we this can be pulled in with that? a.k.a can we fix that one and make it merge-able?

jcvenegas:
@ganeshmaharaj  right, I did not wanted to modify a lot because there was that PR that is looking to fully enable containerd.

On this PR we only add the logic to point to a cri socket, in the containerd PR there the approach used is to duplicate the kubeadm.yml file, I prefer this way.

The containerd PR additionally do the next:

- Script to install containerd from static binaries. Seems that that should be removed, and use containerd provided by Clear Linux. I share the feeling with @egernst  to have scripts to install static binaries, more for development/testing. Probably we need to move  those scripts in other place/repo

- runc-runtimeClass.yaml: I need to confirm if still needed today ( I dont think so, so probably need to remove from the PR as well)
- containerd.conf I think that could be more dynamic in a script to set correct paths to
RuntimeRoot = "/usr/local/bin/kata-runtime"

mythi:
> On this PR we only add the logic to point to a cri socket,

`kubeadm` is able to auto-detect the socket these days. How about we add `kustomize` to manage `kubeadm.yaml` and see if the autodetection works for `kubeadm reset`?

jcvenegas:
@mythi seems that works fine with containerd, I'll need to check with crio 
@ganeshmaharaj @krsna1729 any preference on the method.

Be explicit and allow the user change the default paths or let kubeadm pick the path based en well known paths.

krsna1729:
Yes lets keep it simple. I have been using autodetection for a few days. Seems to be working. If the user wants custom location they can add. Out of the box crio and containerd locations as packaged by clear must work.

jcvenegas:
@krsna1729 thanks for confirm.  I squashed the commits.
@mythi  kubeadm.yaml does not need any custom modification at the moment after this change, so kustomize  not needed for this.

mythi:
@krsna1729 @jcvenegas sounds good!

Note typo in the commit message

jcvenegas:
typo fixed

ganeshmaharaj:
Seems we are all happy with this change. time to hit merge.

PR #98, opened 2019-05-30
calico-node fails to come up cause of all.rp_filter

ganeshmaharaj:
Currently net.ipv4.conf.all.rp_filter is set to 2 in Clear and
calico-node fails to come up unless that value is either 0 or 1.

### DIFF
--- a/clr-k8s-examples/setup_system.sh
+++ b/clr-k8s-examples/setup_system.sh
@@ -23,6 +23,7 @@ sudo mkdir -p /etc/sysctl.d/
 cat <<EOT | sudo bash -c "cat > /etc/sysctl.d/60-k8s.conf"
 net.ipv4.ip_forward=1
 net.ipv4.conf.default.rp_filter=1
+net.ipv4.conf.all.rp_filter=1
 EOT
 sudo systemctl restart systemd-sysctl
 


mcastelino:
> @ganeshmaharaj did this setting change in newer versions of clearlinux? Or kernel defaults changed?

@bryteise any idea why we needed to change kernel defaults.

miguelinux:
mmm ... looks like always be `2` because  `RP_FILTER` is set to `MAXCONF`

https://elixir.bootlin.com/linux/v5.1.5/source/include/linux/inetdevice.h#L99
https://elixir.bootlin.com/linux/v5.1.5/source/include/linux/inetdevice.h#L92

on what version of kernel Does calico-node worked good?

miguelinux:
from: https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt
```
rp_filter - INTEGER
	0 - No source validation.
	1 - Strict mode as defined in RFC3704 Strict Reverse Path
	    Each incoming packet is tested against the FIB and if the interface
	    is not the best reverse path the packet check will fail.
	    By default failed packets are discarded.
	2 - Loose mode as defined in RFC3704 Loose Reverse Path
	    Each incoming packet's source address is also tested against the FIB
	    and if the source address is not reachable via any interface
	    the packet check will fail.

	Current recommended practice in RFC3704 is to enable strict mode
	to prevent IP spoofing from DDos attacks. If using asymmetric routing
	or other complicated routing, then loose mode is recommended.

	The max value from conf/{all,interface}/rp_filter is used
	when doing source validation on the {interface}.

	Default value is 0. Note that some distributions enable it
	in startup scripts.

```

miguelinux:
Did you change the platform from it is running where it has more then one NIC?

krsna1729:
ya i am guessing the check started recently in calico-node?

miguelinux:
systemd 240 introduced that change

miguelinux:
https://github.com/systemd/systemd/blob/master/NEWS#L574

```
      * The "net.ipv4.conf.all.rp_filter" sysctl will now be set to 2 by
          default. This effectively switches the RFC3704 Reverse Path filtering
          from Strict mode to Loose mode. This is more appropriate for hosts
          that have multiple links with routes to the same networks (e.g.
          a client with a Wi-Fi and Ethernet both connected to the internet).


```

PR #100, opened 2019-06-06
WIP: Add support for persistent memory modeling

mcastelino:
Add support for persistent memory in Vagrant. This will allow
us to test pmem-csi plugins as well as other usages of
persistent memory in cloud-native setup

(Somebody check my Rubies)


krsna1729:
how do you plan to handle selinux/apparmor blocking access to /dev/shm?

mcastelino:
@krsna1729 yes that adds a quirk to how users bring up the cluster. Today we tell them just to do vagrant up. Now we will need to add a manual step. This is not ideal. I would not like to wrap the vagrant file in a script. 

This requires elevated privileges to do this.

https://gist.github.com/mcastelino/d7477dd6183244c5fe64c17c9b4c2c88

PR #112, opened 2019-07-09
Versioning: Update clearlinux version to 30270

mcastelino:
Clearlinux release 30270 is now a known good version compatible
with kubernetes.

RuntimeName:  cri-o
RuntimeVersion:  1.14.4
RuntimeApiVersion:  v1alpha1

Kubernetes v1.15.0

runc version 1.0.0-rc5
spec: 1.0.0

systemd 242 (242)
+PAM +AUDIT -SELINUX +IMA -APPARMOR -SMACK -SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ -LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD -IDN2 -IDN -PCRE2 default-hierarchy=legacy

Kernel: 4.19.57-60.lts2018

### DIFF
--- a/clr-k8s-examples/setup_system.sh
+++ b/clr-k8s-examples/setup_system.sh
@@ -7,7 +7,7 @@ ADD_NO_PROXY="10.244.0.0/16,10.96.0.0/12"
 ADD_NO_PROXY+=",$(hostname -I | sed 's/[[:space:]]/,/g')"
 
 #Install kubernetes and crio
-sudo -E swupd repair --picky -m 29880 --force
+sudo -E swupd repair --picky -m 30270 --force
 sudo -E swupd bundle-add cloud-native-basic storage-utils
 
 #Permanently disable swap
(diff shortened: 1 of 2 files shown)


krsna1729:
@mcastelino at what point do we get rid of the pinning. Both in the script and box version

mcastelino:
> @mcastelino at what point do we get rid of the pinning. Both in the script and box version

Once we have the automated CI in place we should stop pinning.

/cc @hnanni @nitinatintel

clr-k8s-examples/setup_system.sh
```diff
@@ -7,7 +7,7 @@ ADD_NO_PROXY="10.244.0.0/16,10.96.0.0/12"
 ADD_NO_PROXY+=",$(hostname -I | sed 's/[[:space:]]/,/g')"
 
 #Install kubernetes and crio
-sudo -E swupd repair --picky -m 29880 --force
```
pohly: Do you also disable swupd auto-update somewhere? Otherwise locking onto a specific version here just helps with the initial installation, later system updates then can still break the Kubernetes cluster (either because of bugs or because kubelet gets updated to a new major version before the admin has a chance to update the control plane).

PR #114, opened 2019-07-12
Adding a script to execute upstream k8s e2e tests on the cloud native stack and took pinning out of setup_system

NitinAtIntel:


clr-k8s-examples/tests/e2e/run_e2e.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo Running e2e tests where spec matches $1
+else
+    echo Running all e2e tests, this will take a long time
+fi
+
+sudo swupd bundle-add go-basic
+
+go get -d k8s.io/kubernetes
+go get -u k8s.io/test-infra/kubetest
+
```
NitinAtIntel: Moved the error on exit options below go get calls as even a successful get of k8s.io/kubernetes results in an error message stating that no go files were found.  However, the repo seems to have been pulled fine and tests run.

clr-k8s-examples/tests/e2e/run_e2e.sh
```diff
@@ -0,0 +1,41 @@
...
+set -o errexit
```
ganeshmaharaj: We would want `errexit` to happen even if we are unable to do a `swupd bundle-add` or `git clone...` right? maybe we should move this to the top of the file?

clr-k8s-examples/tests/e2e/run_e2e.sh
```diff
@@ -0,0 +1,41 @@
...
+set -o pipefail
+set -o nounset
+
+cd /home/clear/go/src/k8s.io/kubernetes
```
ganeshmaharaj: Maybe we can use `GOPATH` variable here? This way the scripts will/can run outside of vagrant setup as well?

clr-k8s-examples/tests/e2e/run_e2e.sh
```diff
@@ -0,0 +1,41 @@
...
+
+PATH=$PATH:/home/clear/go/bin
```
ganeshmaharaj: Maybe use `GOBIN` here?

clr-k8s-examples/tests/e2e/run_e2e.sh
ganeshmaharaj: Does the script fail cause of the message you are getting due to `go get`?

NitinAtIntel: Yes line 21 returns an error even on successful get.  I was researching go get options to ignore the message of no go files found

clr-k8s-examples/tests/e2e/run_e2e.sh
NitinAtIntel: I agree, looking for options to get around line 21 failing on success

clr-k8s-examples/tests/e2e/run_e2e.sh
NitinAtIntel: When I install the go-basic bundle, `GOPATH` is not being set.  Same with `GOBIN`.

ganeshmaharaj: Then probably this would work. 
```
if [ -z $GOPATH ] ; then GOPATH=$HOME/go; fi
if [ -z $GOBIN ] ; then GOBIN=$HOME/go/bin; fi
```

NitinAtIntel: Yes, I will include this in an update to the PR

NitinAtIntel:
Made some changes

clr-k8s-examples/tests/e2e/run_e2e.sh
NitinAtIntel: Using git to get kubernetes instead

clr-k8s-examples/setup_system.sh
```diff
@@ -7,7 +7,6 @@ ADD_NO_PROXY="10.244.0.0/16,10.96.0.0/12"
 ADD_NO_PROXY+=",$(hostname -I | sed 's/[[:space:]]/,/g')"
 
 #Install kubernetes and crio
-sudo -E swupd repair --picky -m 30270 --force
```
ganeshmaharaj: Thanks. We should find a way for the user to specify this. Will add that to the vagrant setup and passs it down to the script from there.

clr-k8s-examples/tests/e2e/run_e2e.sh
```diff
@@ -0,0 +1,65 @@
...
+    cd kubernetes
+    echo Checking status of existing k8s repo clone
+    git status kubernetes
+else
+    echo Cloning upstream k8s repo
+    git clone https://github.com/kubernetes/kubernetes.git
+    cd kubernetes
+fi
+
+PATH=$PATH:$GOBIN
```
ganeshmaharaj: Just a small enhancement. you may want to put this in your `$HOME/.bashrc` so that when you login you can do similar calls. Might help with debug.

clr-k8s-examples/setup_system.sh
ganeshmaharaj: @mcastelino you ok dropping this change? Not sure why we had it hard-coded here, but we should probably come up with a more elegant solution if this is something we will need.

mcastelino: Yes. This was done when kubernetes was broken on clear. The better way to do this is to ensure that the scripts included in the clearlinux cloud-native-basic bundle work with released clearlinux. The clearlinux CI is now validating the the scripts. So this change can now be removed.

krsna1729: @mcastelino Do we want to add back swupd update that was there in place of this?

PR #118, opened 2019-07-19
Move to containerd as the default manager

ganeshmaharaj:
This setup scripts have been using cri-o all this while as the pod
controller/manager system. Recent past cri-o has had some issues with
kata-deploy (A restart of the service will not be able to re-connect
with the existing pods and restart all of them including kata-deploy,
which will hit an endless loop). Moving to containerd as default for
now.

Firecracker cannot be used with a released version of containerd as
there is no default block based snapshotter available today. If you wish
to use cri-o make sure you set `RUNNER=crio` in your environment prior
to using the script.

### DIFF
--- a/clr-k8s-examples/Vagrantfile
+++ b/clr-k8s-examples/Vagrantfile
@@ -20,6 +20,10 @@ $hosts = {}
 $proxy_ip_list = ""
 $driveletters = ('a'..'z').to_a
 $setup_fc = true ? (['true', '1'].include? ENV['SETUP_FC'].to_s) : false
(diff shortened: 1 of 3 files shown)


mcastelino:
@ganeshmaharaj so the current released version of containerd does not have support for device mapper. So that will be reason to have the option here to support cri-o.

Also does the reset script also need changes for containerd?

mythi:
To make the transition complete, the following additional changes are needed:
- website doc updates
- bundle update(s): `cloud-native-basic` installs `cri-o` and `setup_system.sh` adds that bundle only
- BBT updates

Issue #121, opened 2019-07-23
metrics: add a local report generator tool for quick compares


grahamwhaley:
The metrics tests (#119) are not only intended to run under a CI and collect statistics in an ELK style system, but are also very useful for running 'locally' by hand by developers to check for noticeable effects in their local dev work before submitting upstream.

Put together a set of tools to process and present results from local test result files.
Very likely based on the [Kata report generator tooling](https://github.com/kata-containers/tests/tree/master/metrics/report)

PR #122, opened 2019-07-23
metrics: Add the start of the local report generator

grahamwhaley:
Add the first basic report generator code, utilising R, Rmarkdown
and pandoc to do the data processing and pdf report generation.

Fixes: #121

### DIFF
--- a/metrics/report/README.md
+++ b/metrics/report/README.md
@@ -1,0 +1,73 @@
+* [cloud-native-setup metrics report generator](#cloud-native-setup-metrics-report-generator)
+   * [Data gathering](#data-gathering)
+   * [Report generation](#report-generation)
+   * [Debugging and development](#debugging-and-development)
+
+# cloud-native-setup metrics report generator
+
+The files within this directory can be used to generate a 'metrics report'
+for Kubernetes.
+
+The primary workflow consists of two stages:
+
+1) Run the provided report metrics data gathering scripts on the system(s) you wish
+to analyze.
+2) Run the provided report generation script to analyze the data and generate a
+report file.
+
+## Data gathering
+
+Data gathering is provided by the `grabdata.sh` script. When run, this script
+executes a set of tests from the `cloud-native-setup/metrics` directory. The JSON results files
+will be placed into the `cloud-native-setup/metrics/results` directory.
+
+Once the results are generated, create a suitably named subdirectory of
+`tests/metrics/results`, and move the JSON files into it.
+
+Repeat this process if you want to compare multiple sets of results. Note, the
+report generation scripts process all subfolders of `tests/metrics/results` when
+generating the report.
+
+You can restrict the subset of tests run by `grabdata.sh` via its commandline parameters:
+
+| Option | Description |
+| ------ | ----------- |
+| -a | Run all tests (default) |
+| -s | Run the scaling tests |
+| -h | Print this help |
+
+## Report generation
+
+Report generation is provided by the `makereport.sh` script. By default this script 
+processes all subdirectories of the `cloud-native-setup/metrics/results` directory to generate the report.
+To run in the default mode, execute the following:
+
+```sh
+$ ./makereport.sh
+```
+
+The report generation tool uses [Rmarkdown](https://github.com/rstudio/rmarkdown),
+[R](https://www.r-project.org/about.html) and [pandoc](https://pandoc.org/) to produce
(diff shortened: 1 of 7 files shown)


grahamwhaley:
@obedmr @dklyle - first pass. Functional, but has a few rough edges most likely. We can then work in fixing this up in unison with the code in #119.
Note, this PR relies on code that will land with #119 - so we should merge that one first.

Here is an example screenshot of the current output. N.B., this works better with a *large* number of pods, but the default 20 pod run gives us an idea. Also, once we move the data gathering to ms or ns resolution for boot times, the boot graph should be somewhat less 'step function'.

image

metrics/report/report_dockerfile/scaling.R
```diff
@@ -0,0 +1,197 @@
...
+if ( skip_points == 0 ) {
+	cpu_line_plot = cpu_line_plot + geom_point( data=data, aes(count, cpu_idle, colour=testname, group=dataset), alpha=0.3)
+}
+
+# Show how boot time changed
+boot_line_plot <- ggplot() +
+	geom_line( data=data, aes(count, boot_time, colour=testname, group=dataset), alpha=0.2) +
+	geom_smooth( data=data, aes(count, boot_time, colour=testname, group=dataset), se=FALSE, method="loess", size=0.3) +
+	xlab("pods") +
+	ylab("Boot time (s)") +
```
dklyle: I changed this to ms in the PR this depends on.

metrics/report/report_dockerfile/scaling.R
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env Rscript
+# Copyright (c) 2018 Intel Corporation
```
dklyle: year probably needs an update

metrics/report/grabdata.sh
```diff
@@ -0,0 +1,97 @@
...
+   Description:
+        This script gathers a number of metrics for use in the
+        report generation script. Which tests are run can be
+        configured on the commandline. Specifically enabling
+        individual tests will disable the 'all' option, unless
+        'all' is also specified last.
+   Options:
+        -a,         Run all tests (default).
+        -h,         Print this help.
+        -s,         Run the scaling tests.
```
dklyle: Here -s is the specific option, above it's -d

metrics/report/README.md
```diff
@@ -0,0 +1,73 @@
...
+Repeat this process if you want to compare multiple sets of results. Note, the
+report generation scripts process all subfolders of `tests/metrics/results` when
+generating the report.
+
+You can restrict the subset of tests run by `grabdata.sh` via its commandline parameters:
+
+| Option | Description |
+| ------ | ----------- |
+| -a | Run all tests (default) |
+| -d | Run the density tests |
```
dklyle: here the option is -d for density, below it's -s for scaling. This one should be updated to -s I believe.

metrics/report/grabdata.sh
```diff
@@ -0,0 +1,97 @@
...
+		run_scaling
+	fi
+
+	popd
+}
+
+finish() {
+	echo "Now please create a suitably descriptively named subdirectory in"
+	echo "$RESULTS_DIR and copy the .json results files into it before running"
+	echo "this script again."
```
dklyle: Perhaps we could automate this with a timestamp named directory? dir_name=$(date +%s)

metrics/report/README.md
grahamwhaley: Ah, thanks. Yep, I did a s/density/scaling/ but looks like I missed a few...

metrics/report/grabdata.sh
grahamwhaley: Yeah, having this more automated would be nice. I think rather than having this script create a dir and then copying `../results/*.json` into it (which feels a bit error prone and hacky), probably better that we make the results destination dir that the tests themselves write into over-rideable. Right now it is set in `common.bash` like:

```bash
RESULT_DIR="${LIB_DIR}/../results"
```

How about after #119 lands, we update that to more like:

```bash
RESULT_DIR="${RESULT_DIR:-${LIB_DIR}/../results}"
```

and then also update this script to set a `RESULT_DIR` before it invokes the tests themselves. That feels more sustainable...

metrics/report/report_dockerfile/scaling.R
grahamwhaley: done

PR #125, opened 2019-07-30
Add parallel scale test, and cleanups

grahamwhaley:
Add a test to measure scalability of parallel container launch/delete.
To integrate nicely with the existing code base, make a number of cleanups as well.


obedmr:
/me testing locally and then also in Azure CI

obedmr:
@grahamwhaley thanks for all this, it's working really good, I have a some questions or comments: 

if I already run the `metrics/scaling/k8s_scale.sh` and I want to generate the report, is it possible? I tried running the `makereport.sh` from [report](https://github.com/clearlinux/cloud-native-setup/tree/master/metrics/report) dir and it's not generating it, it says that it's not finding the required files.

I found that the `grabdata.sh` script is running the both `k8s_scale.sh` and `k8s_parallel.sh`, shouldnt this be only for grabing pre-generated data?

metrics/scaling/k8s_parallel.sh
```diff
@@ -0,0 +1,293 @@
+#!/bin/bash
```
dklyle: I think we should either combined this with k8s_scale.sh or extract more common tooling to share, one of the scripts is going to get stale quickly. But we can start this way.

metrics/lib/common.bash
```diff
@@ -42,93 +44,6 @@ check_cmds()
 	done
 }
 
-# This function performs a docker pull on the image names
```
dklyle: happy to see all this go!

grahamwhaley:
np @obedmr 
- `grabdata.sh` maybe should be called more correctly `generatedata.sh`. It actually runs the tests to get a new set of data.
- `makereport.sh` looks in subdirs of `results` for data to process - if you don't hand-move the data from `grabdata.sh` into (well named) subdirs, then it will not find anything to process.... the subdirs is because `makereport.sh` can (and mostly is used to) handle multiple sets of data - so you can compare changes etc.

So, fairly normal operation is like:
- grabdata
- mkdir results/run1; mv *.json run1
- grabdata
- mkdir results/run2; mv *.json run2
- makereport.sh

iyswim.

@dklyle did suggest `grabdata` could make a datestamped dir and put the results in it automatically - which, I think we could do pretty easily. I can do that in another PR (and add some better docs) if we want.

metrics/scaling/k8s_parallel.sh
grahamwhaley: I agree. My gut says share the common code via libs, depending on just how similar tests are. A balancing act.
Yeah, I pulled out the api code in this PR, and I expect we will pull out and factor our/share more code as we add more tests.

metrics/scaling/k8s_parallel.sh
```diff
@@ -0,0 +1,293 @@
...
+# sleep and timeout times for k8s actions, in seconds
+wait_time=${wait_time:-30}
+delete_wait_time=${delete_wait_time:-600}
+use_api=${use_api:-yes}
+
+# Set some default metrics env vars
+TEST_ARGS="runtime=${RUNTIME}"
+TEST_NAME="k8s parallel"
+
+get_num_nodes() {
```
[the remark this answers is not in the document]
grahamwhaley: good spot - I forgot to dedup them both! Done.

dklyle:
I think you pushed the wrong branch, the last commit if for your other PR

grahamwhaley:
Oops, had that extra commit in for some local testing, and forgot/didn't spot it was still there! apologies!
pushed.. hope I didn't mess it up..

PR #128, opened 2019-07-31
Allow bootstrap token and master_ip to be customizable

ganeshmaharaj:
This change will allow the create_stack.sh script to be called to init
the k8s cluster with a user generated bootstrap token and a pre-defined
master_ip. This will allow us to automate initial deployment of cluster
with a known list of IPs


ganeshmaharaj:
@jascott1 WDYT?

jascott1:
@ganeshmaharaj Looks good. What do you think about name spacing the external vars? I used "CLRK8S_" as a prefix.

PR #129, opened 2019-08-01
Multinode support

dklyle:
This adds the stats pod via a daemonset. stats are collected for all nodes after each pod launch. The json output structure was altered to support this new data. To support the new structure, nested array support in json.bash was added. This only supports one layer of nesting. I looked at generalized nesting support, but 2D data structures in bash leaves a lot to be desired. 

The reporting in R has also been updated to support displaying cpu and memory for all nodes on a cluster for all tests being compared. The page layout was also changed to be one chart per row, rather than 2 as the charts ended up rather small for the greater detail. In the future, we'll likely want to increase the height of the rows with graphs.

### DIFF
--- a/metrics/lib/json.bash
+++ b/metrics/lib/json.bash
@@ -8,6 +8,7 @@
 
 declare -a json_result_array
 declare -a json_array_array
+declare -a json_nested_array
 
 # Default to dropping the data - in the very rare case that we
 # call 'save' before we have completed 'init' (in the quit/cleanup
@@ -134,6 +135,73 @@ metrics_json_add_fragment() {
 
 	# Place on end of array
 	json_result_array[${#json_result_array[@]}]="$data"
+}
+
+# For building an array in an array element
+metrics_json_start_nested_array() {
+    json_nested_array=()
+}
+
+# Add a (complete) element to the current nested array
+metrics_json_add_nested_array_element() {
+	local data=$1
+
+	# Place on end of array
+	json_nested_array[${#json_nested_array[@]}]="$data"
+}
+
+# Add a fragment to the current nested array element
+metrics_json_add_nested_array_fragment() {
+	local data=$1
+
+	# Place on end of array
+	json_nested_array_fragments[${#json_nested_array_fragments[@]}]="$data"
+}
+
+# Turn the currently registered nested array fragments into an array element
+metrics_json_close_nested_array_element() {
+
+	local maxelem=$(( ${#json_nested_array_fragments[@]} - 1 ))
+	local json="$(cat << EOF
+	{
+		$(for index in $(seq 0 $maxelem); do
+			if (( index != maxelem )); then
+				echo "${json_nested_array_fragments[$index]},"
+			else
+				echo "${json_nested_array_fragments[$index]}"
+			fi
+		done)
+	}
+EOF
+)"
+
+	# And save that to the top level
+	metrics_json_add_nested_array_element "$json"
+
+	# Reset the array fragment array ready for a new one
+	json_nested_array_fragments=()
+}
+
+# Close the current nested array
+metrics_json_end_nested_array() {
+	local name=$1
+
+	local maxelem=$(( ${#json_nested_array[@]} - 1 ))
+	local json="$(cat << EOF
+	"$name": [
+		$(for index in $(seq 0 $maxelem); do
+			if (( index != maxelem )); then
+				echo "${json_nested_array[$index]},"
+			else
+				echo "${json_nested_array[$index]}"
+			fi
+		done)
+	]
+EOF
+)"
+
+	# And save that to the top level
+	metrics_json_add_array_fragment "$json"
 }
 
 # Prepare to collect up array elements
--- a/metrics/report/report_dockerfile/scaling.R
+++ b/metrics/report/report_dockerfile/scaling.R
@@ -18,6 +18,7 @@ testnames=c(
 )
 
 data=c()
+fndata=c()
 stats=c()
 rstats=c()
 rstats_names=c()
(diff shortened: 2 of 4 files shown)


metrics/report/report_dockerfile/scaling.R
```diff
@@ -56,8 +56,8 @@ for (currentdir in resultdirs) {
 
 			testname=datasetname
 
-			cdata=data.frame(avail_gb=as.numeric(fdata$BootResults$mem_free$Result)/(1024*1024))
-			cdata=cbind(cdata, cpu_idle=as.numeric(fdata$BootResults$cpu_idle$Result))
+			cdata=data.frame(avail_gb=as.numeric(fdata$BootResults$node_util$mem_free$Result)/(1024*1024))
+			cdata=cbind(cdata, cpu_idle=as.numeric(fdata$BootResults$node_util$cpu_idle$Result))
```
dklyle: This shouldn't have been in this commit and was completely refactored in the next commit. Please ignore for now.

metrics/report/report_dockerfile/scaling.R
```diff
@@ -139,26 +201,28 @@ mem_line_plot <- ggplot() +
...
 	theme=ttheme(base_size=10),
 	rows=NULL
 	))
 
-# plot how samples varioed over  'time'
-cpu_line_plot <- ggplot() +
-	geom_line( data=data, aes(count, cpu_idle, colour=testname, group=dataset), alpha=0.2) +
-	geom_smooth( data=data, aes(count, cpu_idle, colour=testname, group=dataset), se=FALSE, method="loess", size=0.3) +
+# plot how samples varied over  'time'
+cpu_line_plot <- ggplot(data=fndata, aes(as.numeric(as.character(pod)),
```
dklyle: moving this to the base ggplot removes unnecessary duplication

metrics/lib/json.bash
```diff
@@ -136,6 +137,73 @@ metrics_json_add_fragment() {
...
+			fi
+		done)
+	]
+EOF
+)"
+
+	# And save that to the top level
+	metrics_json_add_array_fragment "$json"
+}
+
```
grahamwhaley: Heh, nested fragments - nice! There have been a couple of times in the past when we have considered adding this (but then figured out workarounds), as, as you say, multi-dimentional data in `bash` is just, well, :boom: mindblowing....

metrics/scaling/k8s_scale.sh
```diff
@@ -39,41 +39,68 @@ TEST_NAME="k8s scaling"
...
+EOF
+    )"
+    metrics_json_add_array_fragment "$launch_json"
+
+    metrics_json_start_nested_array
+
+    # grab pods in the stats daemonset
+    # use 3 for the file descriptor rather than stdin otherwise the sh commands
+    # in the middle will read the rest of stdin
+    while read -u 3 name node; do
```
grahamwhaley: Ouch. and, bash magic ;-)

metrics/report/report_dockerfile/scaling.R
```diff
@@ -126,9 +186,11 @@ mem_stats_plot = suppressWarnings(ggtexttable(data.frame(rstats),
 	))
 
 # plot how samples varied over  'time'
-mem_line_plot <- ggplot() +
-	geom_line( data=data, aes(count, avail_gb, colour=testname, group=dataset), alpha=0.2) +
-	geom_smooth( data=data, aes(count, avail_gb, colour=testname, group=dataset), se=FALSE, method="loess", size=0.3) +
+mem_line_plot <- ggplot(data=fndata, aes(as.numeric(as.character(pod)),
```
grahamwhaley: I see you pass the data at the ggplot() level, not to each graph - which is fine, and the normal ggplot idiom - but, iirc, there was some reason in the past I didn't do that - so, we'll just have to keep our eyes on that ;-)

metrics/report/report_dockerfile/scaling.R
grahamwhaley: yep. as above, iirc, I did have some instance in the past where it did not work out for me (possibly having multiple data sets of variants of plots on the same diagram). But, am fine to dedup where we can :-)

metrics/scaling/k8s_scale.sh
```diff
@@ -39,68 +39,68 @@ TEST_NAME="k8s scaling"
...
+	local mem_free=()
 	info "And grab some stats"
 
-    local pods_json="$(cat << EOF
-            "n_pods": {
-                "Result": ${n_pods},
-                "Units" : "int"
-            }
+	local pods_json="$(cat << EOF
+			"n_pods": {
```
grahamwhaley: heh - it is really hard to try and locally format the json txt in the shell files so they generate pretty json in the final output ;-).
hmm, maybe we could spit the json out via a json pretty printer stream?

grahamwhaley:
This is looking nice :-) Thanks for the effort in moving the stats to a daemonset, improving the json to 2-dimentions, and reworking the R - quite a lot of uptake in a single PR :-)
Having seen the pdf output of a multi-node multi-run plot (thanks), a few thoughts on how we might make some of the more complex plots more readable:
- we could colour-group the test runs together. The default pallette does not handle mutilple multi-entry plots so well. Over in the kata fio code, I used a 'paired' pallette to separate out the two different percentiles I was trying to show. https://github.com/kata-containers/tests/blob/master/metrics/report/report_dockerfile/fio-reads.R#L187
 The pallettes may not scale beyond say two runs, and a few nodes, but it might help in the maybe common case (comparing 2 runs with 3 nodes say?). Maybe we can find something on http://moderndata.plot.ly/create-colorful-graphs-in-r-with-rcolorbrewer-and-plotly/
- You could differentiate the runs with the point shapes as well, maybe by adding a `shape=dataset` to the `geom_point` plots.
- For the nodes, we are plotting both the raw line/points and the `geom_smooth`. This is fine, but doubles the 'noise', so I wonder if we still need both - of if we should do a single smooth of the avg of all nodes for a run for instance?

metrics/scaling/k8s_scale.sh
```diff
@@ -67,6 +67,15 @@ EOF
 	# use 3 for the file descriptor rather than stdin otherwise the sh commands
 	# in the middle will read the rest of stdin
 	while read -u 3 name node; do
+		# look for taint that prevents scheduling
```
dklyle: first part of filtering out NoSchedule nodes from later pod / GB and pod / CPU calculation. Currently master is calculated in even if NoSchedule taint is set.

metrics/scaling/k8s_scale.sh
```diff
@@ -159,6 +169,10 @@ run() {
 	trap cleanup EXIT QUIT KILL
 
 	metrics_json_start_array
+
```
dklyle: grab a baseline on the stats

dklyle:
I'm verifying something on this, please don't merge.

dklyle:
OK, fixed issue with push, feel free to merge. A new unrelated json field was causing issues. Removed for this PR.

metrics/report/report_dockerfile/scaling.R
```diff
@@ -56,38 +57,100 @@ for (currentdir in resultdirs) {
...
+					c3=cbind(row[,3]$Result)/(1024*1024)
+					c4=cbind(rep(i, length(udata$nodename.node)))
+					c5=cbind(rep(testname, length(udata$nodename.node)))
+					# create the new row (which is actually the number of nodes of rows)
+					frow=cbind(c1,c2,c3,c4,c5)
+					fudata=rbind(fudata,frow)
+				}
+			}
+			colnames(fudata)=c("node", "cpu_idle", "mem_free", "pod", "testname")
+			# fudata is considered a vector for some reason so converting it to a data.frame
```
grahamwhaley: I think that is because we initialise it with an empty vector using `=c()` higher up, whereas for other `data.frame` types, we initialise by first assigning them the result of a `data.frame()`, like:

```R
                        udata=data.frame(nodename=fdata$BootResults$node_util)
```
and then we append to those with `cbind()` etc.

Not a problem at all, just noting why it is a vector...

PR #131, opened 2019-08-02
Kustomize all the things

### LINKED ISSUE #2: kustomize: Use kustomize to generate the yaml files
This will allow the use of customized upstream yaml files.

jascott1:
This commit replaces our static manifests with Kustomize
versions and adds scripting to apply them.

Closes #2 


clr-k8s-examples/create_stack.sh
```diff
@@ -20,104 +20,226 @@ $(
...
+    #Create a bare metal load balancer.
+    #kubectl apply -f 6-metal-lb/metallb.yaml
+
+    #The config map should be properly modified to pick a range that can live
+    #on this subnet behind the same gateway (i.e. same L2 domain)
+    #kubectl apply -f 6-metal-lb/example-layer2-config.yaml
 }
 
 function minimal() {
-	cluster_init
```
krsna1729: use shfmt https://github.com/mvdan/sh

jascott1: I formatted the file with the `sh` tool as prescribed.

clr-k8s-examples/create_stack.sh
```diff
@@ -50,53 +58,150 @@ function cluster_init() {
...
 
 function kata() {
-	# Install kata artifacts using kata-deploy
-	kubectl apply -f 8-kata/deploy/kata-rbac.yaml
-	kubectl apply -f 8-kata/deploy/kata-deploy.yaml
-	kubectl apply -f 8-kata/
+	KATA_VER=${1:-1.8.0-kernel-config}
+	KATA_URL="https://github.com/kata-containers/packaging.git"
+	KATA_DIR="8-kata"
+	get_repo "${KATA_URL}" "${KATA_DIR}/overlays/${KATA_VER}"
```
ganeshmaharaj: Can we apply the kata files without actually cloning the repository? Say, `kubectl apply -f https://raw.githubusercontent.com/kata-containers/packaging/1.8.0-kernel-config/kata-deploy/kata-deploy.yaml` or an equivalent of that?

clr-k8s-examples/create_stack.sh
```diff
@@ -50,53 +58,150 @@ function cluster_init() {
...
-		echo "Waiting for Rook CRDs"
-		sleep 2
+	ROOK_VER="${1:-v1.0.3}"
+	ROOK_URL="https://github.com/rook/rook.git"
+	ROOK_DIR=7-rook
+	get_repo "${ROOK_URL}" "${ROOK_DIR}/overlays/${ROOK_VER}"
+	set_repo_version "${ROOK_VER}" "${ROOK_DIR}/overlays/${ROOK_VER}/rook"
+	kubectl apply -k "${ROOK_DIR}/overlays/${ROOK_VER}"
+	# wait for the rook OSDs to run which means rooks should be ready
+	while [[ $(kubectl get po --all-namespaces | grep -e 'osd.*Running.*' -c) -lt 3 ]]; do
```
ganeshmaharaj: When i am doing a single node setup, would this still be 3? just tested it and there are no more than 2 osds coming up. I think rook is resilient enough that we should be able to get going when we see the first osd pop up.

ganeshmaharaj:
We might need to add these two lines to `reset_stack.sh` to make sure we can clean and re-create on these systems.

```
sudo -E bash -c "rm -r /var/lib/etcd/*"
sudo -E bash -c "rm -r /var/lib/etcd"
```

clr-k8s-examples/create_stack.sh
jascott1: We can but it breaks the pattern we are using w kustomize and would have to change back if we need to patch anything. Not sure its worth it, is the advantage just that its quicker?

clr-k8s-examples/create_stack.sh
jascott1: I will check out single node approach.

clr-k8s-examples/create_stack.sh
jascott1: plus you have to maintain a list of files for `kubectl apply` per version and that is already taken care of in new approach w Kustomize.

clr-k8s-examples/create_stack.sh
jascott1: This works and I have updated the wait to be 1 OSD.

jascott1:
@ganeshmaharaj updates complete! Please take a look when you get a chance.

clr-k8s-examples/create_stack.sh
ganeshmaharaj: yeah, was just going for quicker, not adding stuff in a folder etc.. but i guess the alternate is not exactly super future proof either. I am ok with it being what you have for now.

PR #134, opened 2019-08-07
tuning node dataset charting, removing skip_points

dklyle:
This is attempting to improve the information presentation with larger data sets.

I feel that the data points are extremely important and skipping them in larger data sets is problematic. So skip_points was stripped out and points are incorporated into the three graphical charts.

This is becoming too much information for one page and I think my next step will be to split the graphs across pages.

### DIFF
--- a/metrics/report/report_dockerfile/scaling.R
+++ b/metrics/report/report_dockerfile/scaling.R
@@ -26,8 +26,6 @@ rstats_names=c()
 cstats=c()
 cstats_names=c()
 
-skip_points=0	# Shall we draw the points as well as lines on the graphs.
-
 # FIXME GRAHAM - bomb if there are no source dirs?!
 
 for (currentdir in resultdirs) {
@@ -111,12 +109,6 @@ for (currentdir in resultdirs) {
 				c3=cbind(subset(fudata,node==nodename)["cpu_idle"])
 				colnames(c3)=paste(nodename,"_cpu_idle", sep="")
 				cdata=cbind(cdata, c3)
-			}
-
-			# convert ms to seconds
-			# FIXME - we should seq from 0 index
-			if (length(cdata[, "boot_time"]) > 20) {
-				skip_points=1
 			}
 
 			# format the pod data from 2 nested columns in a series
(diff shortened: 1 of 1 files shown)


metrics/report/report_dockerfile/scaling.R
```diff
@@ -113,12 +111,6 @@ for (currentdir in resultdirs) {
 				cdata=cbind(cdata, c3)
 			}
 
-			# convert ms to seconds
```
grahamwhaley: I added the point disabling originally when I was doing test runs of 1000 pods - at which 'point' (pun intended...) all the points merge into each other, and you end up with very thick lines the width of the points....
I'm trying to do a 'large n' run right now (100 pods), to see what this update to point size etc. looks like.

grahamwhaley:
So, I accidentally ran the test with a single node cluster (forgot to re-direct my `kubectl`), and got this in the plot:

```
## Error in dimnames(x) <- dn: length of ’dimnames’ [2] not equal to array extent
```

metrics/report/report_dockerfile/scaling.R
```diff
@@ -229,49 +221,40 @@ mem_line_plot <- ggplot(data=fndata, aes(as.numeric(as.character(pod)),
...
+	xlab("pods") +
 	ylab("System CPU Idle (%)") +
 	ggtitle("System CPU usage") +
 	#ylim(0, NA) + # For big machines, better to not 0-index
 	theme(axis.text.x=element_text(angle=90))
 
-if ( skip_points == 0 ) {
-	cpu_line_plot = cpu_line_plot + geom_point(aes(shape=node), alpha=0.3)
-}
+cpu_line_plot = cpu_line_plot +
```
grahamwhaley: this line left over from a search/delete maybe? Line no longer needed (and is not valid syntax) I think.

metrics/report/report_dockerfile/scaling.R
```diff
@@ -92,7 +90,7 @@ for (currentdir in resultdirs) {
 					fudata=rbind(fudata,frow)
 				}
 			}
-			colnames(fudata)=c("node", "noschedule", "cpu_idle", "mem_free", "pod", "testname")
+			colnames(fudata)=c("node", "noschedule", "cpu_idle", "mem_free", "pod", "testname", "render_point")
```
grahamwhaley: I think this "render_point" addition here is where the array index error comes from - as there is not matching extra cbind into the fudata rows above afaict. I found I could not build a plot of a multi-node run either. I removed the addition here, and now it builds/displays.

grahamwhaley:
I did a run on a 3-node `kind` cluster for 100 pods - looks good, looks like:

image

I'll grab a second set of data and run again now, and will await updates to try it again.

I did have one thought - if we find we are processing just a single set of data (one run), then we could probably switch the colour palette to show individual nodes in different colours - as for a single run plot, that is probably more meaningful and useful than all one colour. Just a thought - it can of course be done later ;-)

dklyle:
Added some color management difference for multi vs single test run and fixed the fact that I pushed an old version last time :o

grahamwhaley:
Thanks, works for me now for both 1 and 2 multi-node datasets, and I see the colours for single run plots.
Merging...

PR #135, opened 2019-08-08
Fixed vagrant issue with proxy

### LINKED ISSUE #133: Vagrant error with proxy
When running this : “**vagrant up --provider=libvirt**”, there is an error returned by vagrant
 
Vagrant failed to initialize at a very early stage:
 
There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
an invalid or undefined variable.
 
Path: /home/khanak/new-setup/cloud-native-setup/clr-k8s-examples/Vagrantfile
Line number: 0
Message: undefined local variable or method `proxy_ip_list'

There is some issue with this line https://github.com/clearlinux/cloud-native-setup/blob/master/clr-k8s-examples/Vagrantfile#L84 .
c.proxy.no_proxy =  (ENV['no_proxy']+"#{proxy_ip_list}" || ENV['NO_PROXY']+"#{proxy_ip_list}" || "localhost,127.0.0.1,172.16.10.10#{proxy_ip_list}")

CraigSterrett:
proxy_ip_list variable had scope issues, used as global variable
initially then as a local variable. Changed all to global.

resolves #133 

Signed-off-by: root <root@csterret-desk1.amr.corp.intel.com>

### DIFF
--- a/clr-k8s-examples/Vagrantfile
+++ b/clr-k8s-examples/Vagrantfile
@@ -60,7 +60,7 @@ Vagrant.configure("2") do |config|
   end
 
   $hosts.each do |vm_name, ip|
-    proxy_ip_list = ("#{proxy_ip_list},#{vm_name},#{ip}")
+    $proxy_ip_list = ("#{$proxy_ip_list},#{vm_name},#{ip}")
   end
 
   $hosts.each do |vm_name, ip|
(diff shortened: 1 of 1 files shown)


ganeshmaharaj:
Thank you so much for catching this. Since moving to chameleonsocks, our testing with proxy has slacked a bit. :)

PR #139, opened 2019-08-12
fixing indexing, start with 0

dklyle:
All graphs were starting with 1. The zero index is the system baseline. So moved to starting with 0.
Fixed the calculations to account for the baseline index, so n-1 as the denominator.

Also accounting for when STEP != 1 in both the graphs and the stats.

### DIFF
--- a/metrics/report/report_dockerfile/scaling.R
+++ b/metrics/report/report_dockerfile/scaling.R
@@ -57,24 +57,27 @@ for (currentdir in resultdirs) {
 			testname=datasetname
 
 			cdata=data.frame(boot_time=as.numeric(fdata$BootResults$launch_time$Result)/1000)
+			cdata=cbind(cdata, num_pods=as.numeric(fdata$BootResults$n_pods$Result))
 
 			# format the utilization data
 			udata=data.frame(nodename=fdata$BootResults$node_util)
 			for (i in seq(length(cdata[, "boot_time"]))) {
+				num_pods=fdata$BootResults$n_pods$Result[i]
+				index=i-1
 				if (i == 1) {
 					# first iteration provide column name for c1
 					c1=cbind(node=udata$nodename.node)
-					c2=cbind(udata$nodename.noschedule)
-					c3=cbind(udata$nodename.cpu_idle$Result)
-					c4=cbind(udata$nodename.mem_free$Result)/(1024*1024)
-					c5=cbind(rep(i, length(udata$nodename.node)))
-					c6=cbind(rep(testname, length(udata$nodename.node)))
+					c2=cbind(noschedule=udata$nodename.noschedule)
+					c3=cbind(cpu_idle=udata$nodename.cpu_idle$Result)
+					c4=cbind(mem_free=udata$nodename.mem_free$Result)/(1024*1024)
+					# using index to make chart start with 0 rather than 1
+					c5=cbind(pod=rep(num_pods, length(udata$nodename.node)))
+					c6=cbind(testname=rep(testname, length(udata$nodename.node)))
 					# declare formatted utility data
 					fudata=cbind(c1,c2,c3,c4,c5,c6)
 				}
 				else {
 					# shift to 0 based indexing
-					index=i-1
 					sindex=(index*4)+1
 					eindex=sindex+3
 					# grab 3 columns for next row bind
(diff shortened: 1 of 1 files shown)


metrics/report/report_dockerfile/scaling.R
```diff
@@ -55,26 +55,29 @@ for (currentdir in resultdirs) {
 			fdata=fdata[[shortname]]
 
 			testname=datasetname
-
+
```
grahamwhaley: `git diff` is showing me trailing whitespace on two lines here I think.

metrics/report/report_dockerfile/scaling.R
```diff
@@ -219,6 +221,12 @@ mem_stats_plot = suppressWarnings(ggtexttable(data.frame(rstats),
 	rows=NULL
 	))
 
+if (step != 1) {
+	xlabel = paste("pods x", step)
```
grahamwhaley: I'm going to have a peek at the output, but I thought we could use the `n_pods$Result` from each entry to get the actual number of pods launched, and then use that on the X axis.
We may have to cast a bit of 'magic' on the X axis at some point, as by default I think it will split it into something like 4 or 10 divisions and number it as such, which makes for some oddities like seeing a '2.5 pods' mark now and then, which just does not make sense ;-)  Last time I looked at this it is slightly painful to get the scales to look good for all our ranges of pods (from say 5 pods to 1000 pods) - but, we can try at some point.

metrics/report/report_dockerfile/scaling.R
dklyle: haven't found a great vim plugin for R yet, so I keep missing these. I have tabs highlighted, but so many tabs, I keep missing the hanging ones.

metrics/report/report_dockerfile/scaling.R
dklyle: I can try to play with that. boot_line_plot uses two data.frames, but I just need to make sure n_pods$Result gets propagated through

grahamwhaley:
I'm probably going to do a bit of debug on some of this tomorrow. I ran a test with a 'kind' stack and some steps, and I see warnings on both the scaling and parallel renders:

Scaling:
```
## Warning in (function (..., row.names = NULL, check.rows = FALSE,
## check.names = TRUE, : row names were found from a short variable and have
## been discarded
## Warning in (function (..., row.names = NULL, check.rows = FALSE,
## check.names = TRUE, : row names were found from a short variable and have
## been discarded
```

Parallel:
```
## Warning in simpleLoess(y, x, w, span, degree = degree, parametric =
## parametric, : span too small. fewer data values than degrees of freedom.
## Warning in simpleLoess(y, x, w, span, degree = degree, parametric =
## parametric, : pseudoinverse used at 0.985
## Warning in simpleLoess(y, x, w, span, degree = degree, parametric =
## parametric, : neighborhood radius 2.015
## Warning in simpleLoess(y, x, w, span, degree = degree, parametric =
## parametric, : reciprocal condition number 0
## Warning in simpleLoess(y, x, w, span, degree = degree, parametric =
## parametric, : There are other near singularities as well. 4.0602
....
```

Both graphs rendered, but the warning/errors come out in the report, and are probably indicative of something we can improve.

Just for ref, the script I ran to do the data grab was:
```bash#!/bin/bash

set -x
rm -f ../results/*.json
runtimes="runc kata-qemu"

for r in $runtimes; do

        # runc is the default, so no 'RUNTIME' setting
        if [ "$r" == "runc" ]; then
                rt=""
        else
                rt=$r
        fi

        #### test the default runtime
        export RUNTIME="$rt"

        ## Grab the scaling 5 pod boot
        export NUM_PODS=50
        export STEP=10
        ../scaling/k8s_scale.sh

        # broke??
        ## Grab the 'time for n' one-shot data
        export NUM_PODS=20
        export STEP=5
        ../scaling/k8s_parallel.sh

        # and store
        mkdir -p "../results/${r}"
        mv ../results/*.json "../results/${r}/"
done
```

grahamwhaley:
The parallel warnings are because my dataset only had 4 or 5 points, and that is probably not enough for the `loess` smoothing line algo to be happy. Maybe I need to set a lower limit to the number of points that invoke the smoothing line - or, find a way to dump the warnings...

dklyle:
Ok, the warning is fixed, it was a data.frame() call that didn't like NAs in the data. So I don't include that row in the data.frame. This creates some indexing fun which I attempted to document in the code. Also using n_pods$Result value now, so the x-axis properly scales. Also using n_pods for the calculations.

metrics/report/report_dockerfile/scaling.R
```diff
@@ -113,44 +116,43 @@ for (currentdir in resultdirs) {
...
+				c5=cbind(dataset=rep(testname, length(pdata$pod_name)))
+				prow=cbind(c1,c2,c3,c4,c5)
+				pudata=rbind(pudata,prow)
 			}
 			# pndata is considered a vector for some reason so converting it to a data.frame
 			pudata=as.data.frame(pudata)
 			pudata$count=as.numeric(as.character(pudata$count))
 			pudata$boot_time=as.numeric(as.character(pudata$boot_time))
 			
-			cdata=cbind(cdata, count=seq_len(length(cdata[, "boot_time"])))
```
dklyle: This was no longer being used, so I removed it.

PR #144, opened 2019-08-14
metrics: parallel plot cleanups

grahamwhaley:
Take into account 'steps' on x-axis plots.
Suppress warnings from geom_smooth.

### DIFF
--- a/metrics/report/report_dockerfile/parallel.R
+++ b/metrics/report/report_dockerfile/parallel.R
@@ -24,7 +24,7 @@ rstats_names=c()
 cstats=c()
 cstats_names=c()
 
-skip_points=0	# Should we draw the points as well as lines on the graphs.
+skip_points_enable_smooth=0	# Should we draw the points as well as lines on the graphs.
 
 for (currentdir in resultdirs) {
 	dirstats=c()
@@ -60,8 +60,10 @@ for (currentdir in resultdirs) {
 
 			# If we have more than 20 items to draw, then do not draw the points on
 			# the graphs, as they are then too noisy to read.
+			# But, do draw the smoothed lines to help read the now dense and potentially
+			# noisy graphs.
 			if (length(cdata[, "boot_time"]) > 20) {
-				skip_points=1
+				skip_points_enable_smooth=1
 			}
 
 			cdata=cbind(cdata, testname=rep(testname, length(cdata[, "boot_time"]) ))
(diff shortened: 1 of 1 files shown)


obedmr:
@grahamwhaley,  I had tested more the `k8s_scale.sh` and it's working, but when I ran the `k8s_parallel.sh` it seems like failing. Is there a missing dependency on my host? I'm digging into the code to see what it could be.

```
➜  scaling git:(pr144) ✗ wait_time=600 NUM_PODS=20 settle_time=0 grace=0 use_api=yes STEP=100   ./k8s_parallel.sh
INFO: Initialising
INFO: Checking k8s accessible
INFO: 2 k8s nodes found
starting kubectl proxy
INFO: Warming up
INFO: Applying warmup pod
Starting to serve on 127.0.0.1:8090
error: error parsing /home/onmunoz/dev/k8s/cloud-native-setup/metrics/scaling/generated.yaml: error converting YAML to JSON: yaml: line 27: found character that cannot start any token
INFO: Cleaning up
Error from server (NotFound): deployments.extensions "busybox" not found
INFO: deployment has terminated
cleaning up kubectl proxy
```

obedmr:
This is my `generated.yaml` file

```
apiVersion: apps/v1

kind: Deployment
metadata:
  labels:
    run: busybox
  name: busybox
spec:
  replicas:
  selector:
    matchLabels:
      run: busybox
  template:
    metadata:
      labels:
        run: busybox
        magiclabel: parallel
    spec:
      terminationGracePeriodSeconds: @GRACE@
      automountServiceAccountToken: false
      containers:
      - name: bb
        image: busybox
        command:
            - "tail"
            - "-f"
            - "/dev/null"
        stdin: true
        tty: true
      restartPolicy: Always
```

grahamwhaley:
@obedmr - yeah, I will likely need to rebase over #142 now that has landed - the 'grace' issue got fixed in that set of commits....

Let me pull/rebase/repush...

metrics/report/report_dockerfile/parallel.R
```diff
@@ -114,11 +111,14 @@ if ( skip_points == 0 ) {
 
 # See https://www.r-bloggers.com/ggplot2-easy-way-to-mix-multiple-graphs-on-the-same-page/ for
 # excellent examples
-master_plot = grid.arrange(
+# Explicitly suppress warnings during the generation of the plot, as geom_smooth can generate a mass
```
dklyle: I'm wondering if suppressing warnings here is a good idea? We can just acknowledge that on runs of < 10 we'll get a warning, or we can remove the smoothing. Just afraid of what else it might mask.

metrics/report/report_dockerfile/parallel.R
```diff
@@ -57,37 +56,35 @@ for (currentdir in resultdirs) {
...
 		}
 	}
 }
 
 # Show how boot time changed
 boot_line_plot <- ggplot() +
-	geom_line( data=data, aes(count, boot_time, colour=testname, group=dataset), alpha=0.2) +
-	geom_smooth( data=data, aes(count, boot_time, colour=testname, group=dataset), se=FALSE, method="loess", size=0.3) +
+	geom_line( data=data, aes(npod, boot_time, colour=testname, group=dataset), alpha=0.2) +
+	geom_smooth( data=data, aes(npod, boot_time, colour=testname, group=dataset), se=FALSE, method="loess", size=0.3) +
```
dklyle: putting the aes portion in ggplot would remove the need for duplication. If we remove the smoothing, it wouldn't matter. Not a blocking comment, just a note

grahamwhaley:
done - that should have fixed it @obedmr I think.

metrics/report/report_dockerfile/parallel.R
grahamwhaley: Agreed. I contemplated it briefly. Sadly there seems no trivial way to calculate if/when warnings will get spat out, and the only clue I could find to some sort of boundary was '1000 points', which I think is more when it changes algorithm rather than spit out warnings.
Let's pick a number and I'll add the if/else. I'm thinking >20 data points maybe? (at the point where the view starts to get dense, and a smoothed line might be useful).

metrics/report/report_dockerfile/parallel.R
grahamwhaley: Ack. I'll try to remember to clean and dedup the aes's during further edits. I wish I could remember why I split them out in the first place... but it was a long time ago in a project far far away...

dklyle: you can pass different things to aes, maybe that was the reason?

grahamwhaley:
DON'T MERGE - taking care of @dklyle feedback..

grahamwhaley:
OK, repushed and dropping DNM.
I've cleaned up the dots/smoothing and dedup'd the aes() code.
I also fixed up some of the graph titles to be more accurate etc. ptal. thx!

PR #151, opened 2019-08-20
[RFC]: metrics: report: Add debug, symbolise scaling node data extraction

grahamwhaley:
Two semi related commits (yeah, really I should have split to two pull requests....)

- When using the makereport script in debug mode, map the scripts dir in from the host so you can 'live edit' the scripts from the host and test them in the container (and, yes, I should add that to the documentation).
- Update the node_util data parsing to use symbolic lookups rather than using array position knowledge, to detach the code from knowing the physical layout of the data - this will allow us to add more data into the structure and have the parsing still work, as long as we don't change the data symbolic names.

### DIFF
--- a/metrics/report/report_dockerfile/scaling.R
+++ b/metrics/report/report_dockerfile/scaling.R
@@ -59,43 +59,32 @@ for (currentdir in resultdirs) {
 			cdata=data.frame(boot_time=as.numeric(fdata$BootResults$launch_time$Result)/1000)
 			cdata=cbind(cdata, num_pods=as.numeric(fdata$BootResults$n_pods$Result))
 
-			# format the utilization data
-			udata=data.frame(nodename=fdata$BootResults$node_util)
-			for (i in seq(length(cdata[, "boot_time"]))) {
-				num_pods=fdata$BootResults$n_pods$Result[i]
-				index=i-1
-				if (i == 1) {
-					# first iteration provide column name for c1
-					c1=cbind(node=udata$nodename.node)
-					c2=cbind(noschedule=udata$nodename.noschedule)
-					c3=cbind(cpu_idle=udata$nodename.cpu_idle$Result)
-					c4=cbind(mem_free=udata$nodename.mem_free$Result)/(1024*1024)
-					# using index to make chart start with 0 rather than 1
-					c5=cbind(pod=rep(num_pods, length(udata$nodename.node)))
-					c6=cbind(testname=rep(testname, length(udata$nodename.node)))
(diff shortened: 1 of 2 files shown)


grahamwhaley:
@dklyle - an RFC, as I'd like your eyes on this (as I am modifying your code), and I should also update the README (which I'll do tomorrow).
The debug item is useful though ;-)

metrics/report/report_dockerfile/scaling.R
```diff
@@ -59,43 +59,32 @@ for (currentdir in resultdirs) {
...
+			# Gather up each 'column' from the dataframe - one per run instance.
+			# The length(node) parts are as each entry has an n-node array of
+			# results, so each iteration can create more than one new row in the
+			# final table.
+			fudata=c()
+			furoot=fdata$BootResults
+			for (n in seq(length(furoot$n_pods$Result))) {
+				num_pods=furoot$n_pods$Result[[n]]
+				u=furoot$node_util[[n]]
+				# first iteration provide column name for c1
```
grahamwhaley: Hmm, maybe I need to nuke that comment line, and/or add back in some column naming (which can be done at the end - assigning a c(names) to colnames(fudata) iirc).

metrics/report/report_dockerfile/scaling.R
```diff
@@ -59,43 +59,32 @@ for (currentdir in resultdirs) {
...
+				c1=cbind(node=u$node)
+				c2=cbind(noschedule=u$noschedule)
+				c3=cbind(cpu_idle=u$cpu_idle$Result)
+				c4=cbind(mem_free=u$mem_free$Result)/(1024*1024)
+				# using index to make chart start with 0 rather than 1
```
grahamwhaley: and.... @dklyle - let me know if I've messed up the 0 indexing... the graph I think looked OK (both old and new versions indexed from 0?). I'll check again though. heh, knew this should all be an RFC ;-)

dklyle: I did not get to try this today, but it looks reasonable to me. That comment should have been removed with my last fix. Feel free to remove it.

metrics/report/report_dockerfile/scaling.R
dklyle: yes, can kill this comment

dklyle:
This looks good to me. I'll approve when you say you're ready.

grahamwhaley:
I'm going to close/deprecate this one - it inspired me to do further updates - new PR imminent...

Issue #148, opened 2019-08-15
podSubnet in kubeadm.yaml ignored


dklyle:
The value set specified for the podSubnet is: 
podSubnet: 10.244.0.0/16

When the cluster comes up via `create_stack.sh minimal` and even when running `kubeadm init --config ./kubeadm.yaml`, the result in `kubectl describe pod <leader-node-name>` gives:
PodCIDR:                     10.244.0.0/24

When the cluster is up with canal configured, the max pod count is 256 which matches 10.244.0.0/24.

NitinAtIntel:
I can take a look at this

krsna1729:
@dklyle in the above command did you mean `kubectl describe node <leader-node-name>`?

If so the behavior is correct. The kubeadm.yaml is specifying the subnet for pods across the entire cluster. This gets divided by into /24 subnets pool per node

dklyle:
@krsna1729 you are correct, sorry for the noise. I missed setting 

```
controllerManager:
  extraArgs:
    node-cidr-mask-size: "20"
```

in kubeadm.yaml.

I will say I could not find any upstream K8s documentation that describes the behavior as you described it. Thanks for the clarification. I will close this now.

krsna1729:
@dklyle now worries. Yup had dig through quite a bit the first time i had to support more that 254 pods. @grahamwhaley needed it too for his tests.

NitinAtIntel:
Thanks folks, we're just trying to slam through these issues so are grabbing whatever we think is relevant :) - so thanks for closing and sorry for churn on assignment.

PR #152, opened 2019-08-20
Adding HIGH_POD_COUNT configuration

dklyle:
This script configures a node in the cluster for pod scaling
testing. This configuration is necessary for pushing beyond the
110 pod max default in Kubernetes.


krsna1729:
This would mostly go into `system_setup.sh` which doesnt need knowledge of kustomize.

As far as changes to `/var/lib/kubelet/config.yaml` are concerned they will be captured in `kubeadm.yaml` under `kind: KubeletConfiguration` to propagate to all nodes joining the cluster.
No knowledge of kustomize required here too. Although it seems like kustomize support for kubeadm is coming https://github.com/kubernetes/kubernetes/pull/80905 so we may want to leverage that.

dklyle:
The question would be if we want a large scale config by default. I'm not aware of any downsides, but I can't say that definitely.

krsna1729:
I think it will be some arg to gate it if we were to move there.

dklyle:
I'm willing to rework for clr-k8s-examples if that is the consensus opinion.

krsna1729:
@ganeshmaharaj? Seems like there are env vars today for both setup_system.sh and create_stack.sh. Should one more be added e.g., `HIGH_POD_COUNT=true` to use the logic in the PR?

metrics/config/node-config.sh
```diff
@@ -0,0 +1,47 @@
...
+LimitCORE=1048576
+TimeoutStartSec=0
+MemoryLimit=infinity
+EOT
+
+sudo systemctl daemon-reload
+sudo systemctl restart kubelet
+
+sudo mkdir -p /etc/systemd/system/containerd.service.d/
+cat <<EOT | sudo bash -c "cat > /etc/systemd/system/containerd.service.d/limits.conf"
```
obedmr: It's looking good to me. The only thing is that if we decide or not to move this to the clr-k8s-examples area, we may use the $RUNNER var for the container engine service config, with this we can continue providing diferent container engines support. @krsna1729  @ganeshmaharaj, @dklyle @grahamwhaley,  any thoughts?

dklyle:
actually related to #148 

I'd like to add to add also per `HIGH_POD_COUNT=true` toggle the following to kubeadm.yaml 

```
controllerManager:
  extraArgs:
    node-cidr-mask-size: "20"
```

dklyle:
Note to self, need to add `swupd bundle add bc`

clr-k8s-examples/setup_system.sh
```diff
@@ -79,6 +80,55 @@ function setup_hosts() {
...
+	if [ "$RUNNER" == "containerd" ]; then
+		sudo mkdir -p /etc/systemd/system/containerd.service.d
+		write_limits_conf "/etc/systemd/system/containerd.service.d/limits.conf"
+	fi
+	if [ "$RUNNER" == "crio" ]; then
+		sudo mkdir -p /etc/systemd/system/crio.service.d
+		write_limits_conf "/etc/systemd/system/crio.service.d/limits.conf"
+	fi
+
+	# increase limits in kubelet
```
krsna1729: is /var/lib/kubelet/config.yaml created before cluster init? is it better to propagate via kubeadm.yaml?

ganeshmaharaj: @dklyle i am already doing monkeying if this kind here. https://github.com/clearlinux/cloud-native-setup/blob/master/clr-k8s-examples/create_stack.sh#L35 😜. I am sure adding `HIGH_POD_COUNT` to that won't hurt us.

dklyle: @krsna1729 I assumed that it was created on install of kubelet. I can verify that, or ...

@ganeshmaharaj my only concern here was having to set the environment variable for two scripts rather than just the one. So I tried to centralize all the work in one script. If we prefer to move this logic to be co-located in create_stack.sh, I can do that.

clr-k8s-examples/create_stack.sh
```diff
@@ -63,6 +74,12 @@ function cluster_init() {
 	if [ "$(kubectl get nodes | wc -l)" -eq 2 ]; then
 		kubectl taint nodes --all node-role.kubernetes.io/master-
 	fi
+
+	if [[ -n "${HIGH_POD_COUNT}" ]]; then
+		# increase limits in kubelet
+		sudo sed -i 's/^maxPods\:.*/maxPods\: 5000/' /var/lib/kubelet/config.yaml
```
krsna1729: i was expecting these two would be updated in kubeadm.yaml, in kubelet config section: https://github.com/clearlinux/cloud-native-setup/blob/044146703f35342bcf58025624128a2eb635ee29/clr-k8s-examples/kubeadm.yaml#L4-L13

Current version of PR only updates master node

clr-k8s-examples/setup_system.sh
```diff
@@ -79,6 +80,44 @@ function setup_hosts() {
...
 
+# write increased limits to specified file
+function write_limits_conf() {
+	cat <<EOT | sudo bash -c "cat > $1"
+[Service]
+LimitNOFILE=1048576
+LimitNPROC=1048576
+LimitCORE=1048576
+TimeoutStartSec=0
+MemoryLimit=infinity
```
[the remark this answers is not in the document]
obedmr: Since this is something that is coming soon into the ClearLinux's containerd, we can omit this request. Thanks @dklyle, let's merge this.

PR #153, opened 2019-08-22
Remove pinning on clearlinux version

ganeshmaharaj:
While we previously had a few instabilities that pegged the version of
Clear, it is time we left it to be the bleeding edge to make sure that
the cloud-native-setup catches and reacts to latest changes on Clear.

### DIFF
--- a/clr-k8s-examples/Vagrantfile
+++ b/clr-k8s-examples/Vagrantfile
@@ -47,7 +47,6 @@ Vagrant.configure("2") do |config|
   # Every Vagrant development environment requires a box. You can search for
   # boxes at https://vagrantcloud.com/search.
   config.vm.box = $box
-  config.vm.box_version = "30260"
 
   # Mount the current dir at home folder instead of default
   config.vm.synced_folder './', '/vagrant', disabled: true


ganeshmaharaj:
/cc @mcastelino @krsna1729 @NitinAtIntel 
@ahsan518 This should work around the issue we had where `swupd update` was failing. We need to be on a later version of Clear for the newer swupd client fixes.

krsna1729:
The update happens in the system setup script. This only says the box to use. The thinking was that for vagrant users this will reflect the last box the setup was tested on.

CraigSterrett:
I didn't see this PR, I actually had the same issue and had a different fix. Allowing an override via environment variable and if no value is set then do not pass a box version. In a perfect world this would grab the latest box, which it does if you don't already have a version on your system. Unfortunately if you have a version on your system, it grabs that one
So not a perfect solution by any means

root@csterret-desk1.amr.corp.intel.com~/git/cloud-native-setup/clr-k8s-examples # grep -i box Vagrantfile
$box = "AntonioMeireles/ClearLinux"
$box_ver = (ENV['CLEAR_VBOX_VER'])
  # Every Vagrant development environment requires a box. You can search for
  # boxes at https://vagrantcloud.com/search.
  config.vm.box = $box
  config.vm.box_version = $box_ver

ganeshmaharaj:
@krsna1729 there is an issue with swupd in one of the later versions of clear where a formatbump happened where swupd will not be able to re-exec itself and fails with 255. That issue hits when we use an old box here and our setup will always fail right after the first swupd update. Hence moving to the latest version would help in the sense our base box is the later version. Sure, i can do update the version number itself, but i thought his hard-coding does not sit well.

@CraigSterrett I see you have an alternate to this patch and you have closed this PR. Is that change in the pipe now? Can we get that merged?

CraigSterrett:
Oops, Didn't mean to close this PR

PR #154, opened 2019-08-22
[RFC] Tidy R for report scaling page

grahamwhaley:
Restructure the scaling.R to accomodate future JSON data shape changes.
Also:
 - make the report dockerfile debug mode nicer
 - add the boot stats table
 - add some rolling data to the JSON, to ease later visualisation

**NOTE** there is still a little to do here, hence the RFC:
 - I think I forgot to add the `pod_per_gb` rolling data to the JSON
 - I'm not sure the colours for the boot chart multi nodes are right - we should stare hard at that.


metrics/report/report_dockerfile/tidy_scaling.R
```diff
@@ -0,0 +1,261 @@
...
+			# but the only other option would be to try and show every node separately in the
+			# table.
+
+			# Get a list of all the nodes
+			nodes=unique(local_nodedata$node)
+
+			memtotal=0
+			cputotal=0
+			# Calculate per-node totals, and tot them up to a global total.
+			for (n in nodes) {
```
dklyle: There should be a bypass for CPU and memory if noschedule == true

metrics/report/report_dockerfile/tidy_scaling.R
```diff
@@ -0,0 +1,261 @@
...
+			local_cpus = c(
+				"Test"=testname,
+				"n"=fdata$Config$NUM_PODS,
+				"Tot_CPU"=round(cputotal, 3),
+				"avg_CPU"=round(cputotal/fdata$Config$NUM_PODS, 4)
+			)
+			cpustats=rbind(cpustats, local_cpus)
+
+			# launch (boot) stats
+			local_boots = c(
```
dklyle: I would have really like to have this today :) This is a great addition.

metrics/scaling/k8s_scale.sh
```diff
@@ -156,6 +171,26 @@ EOF
...
+		local pods_per_gb=$(bc -l <<< "scale=2; ($total_mem_used/1024) / $n_pods")
+	fi
+	local mem_json="$(cat << EOF
+			"memory": {
+				"consumed": {
+					"Result": ${total_mem_used},
+					"Units": "Kb"
+				},
+				"pods_per_gb": {
+					"Result": ${pods_per_gb}
```
dklyle: The value for "Result": was not being populated when I tried it. I set them all to 0 so I could see the results.
The key and colon are added, but not the ${pods_per_gb}. I believe it was set for the first iteration, but none after.

So the don't divide by 0 case worked, but the else was not populating the value.

Just checked, bc is not installed. I'll have to add that to my config patch.

metrics/report/report_dockerfile/tidy_scaling.R
grahamwhaley: adding...

metrics/scaling/k8s_scale.sh
grahamwhaley: Looking into the gb/pod issue.
I've added a call to `check_cmds` (from our lib file) during initialisation to pick up a missing `bc` and report it.
I think it was the missing `bc` that caused the empty results fields, so hopefully that is fixed now..

metrics/scaling/k8s_scale.sh
```diff
@@ -101,6 +104,14 @@ EOF
 		cpu_idle=${cpu_idle:-0}
 		mem_free=${mem_free:-0}
 
+		# If this is the 0 node instance, store away the base memory value
+		if [ $n_pods -eq 0 ]; then
+			node_basemem[$node]=$mem_free
+		fi
+
+		local mem_used=$((node_basemem[$node]-mem_free))
+		total_mem_used=$((total_mem_used+mem_used))
```
grahamwhaley: @dklyle - do you think I should be skipping the `noschedule` nodes here in the total mem used calculation (which is then used in the pods/gb calculation) - I suspect I should....

grahamwhaley:
@dklyle - updated and pushed. I added updates as extra commits, as they were self contained enough, and it should make re-reviewing easier.

metrics/scaling/k8s_scale.sh
dklyle: yes, any of the stats should exclude the noschedule nodes. If the master node becomes a point of interest, I think that would be a separate stats table anyway.

metrics/scaling/k8s_scale.sh
dklyle: yes it was bc, I should have added that as multiple comments rather than editing the same one. My voyage of discovery wasn't clear.

metrics/scaling/k8s_scale.sh
dklyle: probably waiting on my reply, but this is not addressed in the newer commits.

metrics/report/README.md
```diff
@@ -60,14 +60,18 @@ directory.
...
 To aid in script development and debugging, the `makereport.sh` script offers a debug
 facility via the `-d` command line option. Using this option will place you into a `bash`
-shell within the running `Dockerfile` image used to generate the report. From there you
-can examine the Docker image environment, and execute the generation scripts. E.g., to
-test the `scaling.R` script, you can execute:
+shell within the running `Dockerfile` image used to generate the report, whilst also
+mapping your host side `R` scripts from the `report_dockerfile` subdirectory into the
+container, thus facilitating a 'live' edit/reload/run development cycle.
+From there you can examine the Docker image environment, and execute the generation scripts.
+E.g., to test the `scaling.R` script, you can execute:
```
dklyle: scaling.R is deprecated need to change to new R filename.

metrics/report/README.md
```diff
@@ -60,14 +60,18 @@ directory.
...
 
 ```bash
 $ makereport.sh -d
 # R
 > source('/inputdir/Env.R')
 > source('/scripts/scaling.R')
+## Edit script on host, and re-load/run...
+> source('/scripts/scaling.R')
```
dklyle: change here too.

metrics/report/README.md
grahamwhaley: done

metrics/scaling/k8s_scale.sh
grahamwhaley: nah - I only added it a few hours ago ;-)

Issue #136, opened 2019-08-08
Vagrant behind proxy needs no_proxy for CNI IPs


CraigSterrett:
When Vagrant spins up nodes, it needs to add to the no_proxy setting the IP range being used by the CNI. Currently that is 192.168.0.0/16. Without that the worker nodes will not be able to reach the control plane.

PR #155, opened 2019-08-22
Fix no_proxy issue and remove hardcoded box version

CraigSterrett:
Fixed an issue where no_proxy was not getting set to include the
CIDR being used by the CNI. This resulted in kubeadm join command
trying to go out through the proxy server.
Also removed hardcoded vagrant box version. This change will now
allow the override of the version, or if no override is passed in
it will take the version on the system, or download the latest version.

Closes issue #136

### DIFF
--- a/clr-k8s-examples/Vagrantfile
+++ b/clr-k8s-examples/Vagrantfile
@@ -13,8 +13,10 @@ $disks = 2
 $disk_prefix = File.basename(File.dirname(__FILE__), "/")
 $disk_size = "10G"
 $box = "AntonioMeireles/ClearLinux"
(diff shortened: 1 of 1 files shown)


NitinAtIntel:
Super small thing but did you see the commit was made by root? :)

krsna1729:
1. Split into separate PRs
2. Re: proxy - guessing this does not help bare-metal/non-vagrant environments. We want use of vagrant magic to be as minimal as possible. Its only there to provide an infra to test. Lets move the appending of NO_PROXY into system_setup?

CraigSterrett:
OK will split

PR #157, opened 2019-08-22
Support for Virtualbox

ahsan518:
This change adds the env variable for the support for
virtualbox.


ganeshmaharaj:
@ahsan518 could you add a small commit message explanation as to why this change is needed and unique to just virtualbox? This way anyone who tries to understand this change would know why.

Issue #101, opened 2019-06-17
kubelet fails to start


pohly:
I tried `vagrant up --provider=virtualbox` today. It updated to Clear Linux 29940. Then it failed with kubelet refusing to start:
```
    clr-01: Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /usr/lib/systemd/system/kubelet.service.
    clr-01: Created symlink /etc/systemd/system/multi-user.target.wants/crio.service → /usr/lib/systemd/system/crio.service.
    clr-01: Job for kubelet.service failed because the control process exited with error code.
    clr-01: See "systemctl status kubelet.service" and "journalctl -xe" for details.
```

Logging in and checking the journal shows:
```
Jun 17 14:27:20 clr-01 kubelet-version-check.sh[4166]: /usr/bin/kubelet-version-check.sh: line 8: /var/lib/kubelet/kubelet_version: No such file or directory
Jun 17 14:27:20 clr-01 kubelet[4165]: F0617 14:27:20.140206    4165 server.go:194] failed to load Kubelet config file /var/lib/kubelet/config.yaml, error failed to read kub>
```

`/var/lib/kubelet/config.yaml` does not exist.

mcastelino:
@pohly we do not support virtualbox as it does not support all the features needed. Can you use libvirt?

krsna1729:
Do we know if the issue is virtualbox or clearlinux version?

krsna1729:
Since the only hypervisor specific stuff is this https://github.com/clearlinux/cloud-native-setup/blob/0e280f40fa21612f436b7f8d9257d3f850aba9c8/clr-k8s-examples/Vagrantfile#L65-L73

cluster init should ideally work unless i am missing something.

Kata for sure wont work since VBox doesnt support nested

krsna1729:
We could move the cpu and memory out to be agnostic i think

mcastelino:
Looks like we have an issue in Clearlinux itself

```
kubectl run -i --tty busybox --image=busybox --restart=Never -- sh
```

```
Error attaching, falling back to logs: error dialing backend: EOF
Error from server: Get https://192.52.100.12:10250/containerLogs/default/busybox/busybox: net/http: TLS
```

pohly:
> we do not support virtualbox as it does not support all the features needed.

[...]

PR #159, opened 2019-08-23
Support for Virtualbox

ahsan518:
This change adds the env variable for the support for
virtualbox to support kubernetes communication between
master and worker nodes. Since KVM gives default ip
address to the spinned up VM's, whereas virtualbox uses
same ip address for all, the kubelet was not functioning
, this change fixes it and assigns the ip address to env
variable.

### DIFF
--- a/clr-k8s-examples/Vagrantfile
+++ b/clr-k8s-examples/Vagrantfile
@@ -85,6 +85,9 @@ Vagrant.configure("2") do |config|
           c.proxy.no_proxy =  (ENV['no_proxy']+"#{$proxy_ip_list}" || ENV['NO_PROXY']+"#{$proxy_ip_list}" || "localhost,127.0.0.1,172.16.10.10#{$proxy_ip_list}")
         end
       end
+      c.vm.provider :virtualbox do |_, override|
+        override.vm.provision "shell", privileged: true, inline: "sudo mkdir -p /etc/profile.d; echo export MASTER_IP=#{$hosts["clr-01"]} > /etc/profile.d/cnsetup.sh"
+      end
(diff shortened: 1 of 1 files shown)


clr-k8s-examples/Vagrantfile
```diff
@@ -85,6 +85,9 @@ Vagrant.configure("2") do |config|
           c.proxy.no_proxy =  (ENV['no_proxy']+"#{$proxy_ip_list}" || ENV['NO_PROXY']+"#{$proxy_ip_list}" || "localhost,127.0.0.1,172.16.10.10#{$proxy_ip_list}")
         end
       end
+      c.vm.provider :virtualbox do |_, override|
+        override.vm.provision "shell", privileged: true, inline: "sudo mkdir -p /etc/profile.d; echo export MASTER_IP=#{$hosts["clr-01"]} > /etc/profile.d/cnsetup.sh"
```
krsna1729: privileged: true and sudo needed?

ganeshmaharaj:
/cc @CraigSterrett @jascott1 

@pohly this change should help you get past your setup issue referenced at #101

clr-k8s-examples/Vagrantfile
ganeshmaharaj: @krsna1729 yes, cause we are stuffing it inside `/etc/` and it won't let you create a folder there as non-root.

pohly:
I can bring up a four node cluster with `vagrant up --provider=virtualbox` and this PR.  I have not tried whether `./create_stack.sh all` works completely (it seems so, I just didn't join any other nodes and then aborted).

PR #161, opened 2019-08-27
Adding default vagrant CIDR to no_proxy

CraigSterrett:
Vagrant-lbvirt creates a virtual network with a default network CIDR of
192.168.121.0/24. Adding this default CIDR to the no_proxy configuration
in the Vagrant file allows the Vagrant VM's to talk back to the K8s
master.

Closes issue #136

Signed-off-by: craigsterrett <craig.sterrett@intel.com>


CraigSterrett:
Justin found another bug in the overall proxy settings. If you system has http_proxy defined but did not have no_proxy defined then the Vagrant file blew up. I will submit a patch

PR #162, opened 2019-08-28
metrics: json: use tr for newline/whitespace conversion

grahamwhaley:
Use `tr` rather than `sed` for doing the line/whitespace conversions
on the JSON stream, if asked. Slightly more efficient, and we also drop
the wrapping ""'s, to allow the data to land in logstash without error.


grahamwhaley:
@dklyle @obedmr - when moving my local ELK stack from an old beats setup to using logstash to filter, I needed to add/modify this to avoid 'JSON format' errors in logstash. Happens to be slightly more efficient as well I suspect.

PR #164, opened 2019-08-28
Remove possible hyphens from array index

### LINKED ISSUE #163: [metrics] hyphens in system names cause indexing problems
https://github.com/clearlinux/cloud-native-setup/blob/master/metrics/scaling/k8s_scale.sh#L109 is specifically the issue, a hostname like "node-typeA-1" is evaluated as 0-0-1 which thankfully throws an error as it tried to index -1

dklyle:
bash arrays don't handle hyphens in the index as it treats
a name like "azure-east-1" as 0-0-1, so remove any hyphens in the
node name, will work as long as hyphens aren't the only difference
in node names

Fixes #163

### DIFF
--- a/metrics/scaling/k8s_scale.sh
+++ b/metrics/scaling/k8s_scale.sh
@@ -36,7 +36,7 @@ TEST_ARGS="runtime=${RUNTIME}"
 TEST_NAME="k8s scaling"
 
 declare -a new_pods
-declare -a node_basemem
+declare -A node_basemem
 
 # $1 is the launch time in seconds this pod/container took to start up.
 # $2 is the number of pod/containers under test


metrics/scaling/k8s_scale.sh
```diff
@@ -104,12 +104,17 @@ EOF
...
 
+		# bash arrays don't handle hyphens in the index as it treats
+		# a name like "azure-east-1" as 0-0-1, so remove any hyphens in the
+		# name, will work as long as hyphens aren't the only difference
+		node_index=$(tr -d [=-=] <<< $node)
+
 		# If this is the 0 node instance, store away the base memory value
 		if [ $n_pods -eq 0 ]; then
-			node_basemem[$node]=$mem_free
+			node_basemem[$node_index]=$mem_free
```
grahamwhaley: Hmm, I wonder if there is a deeper root cause or an alternative method here... I had a play locally (as bash associative arrays are not exactly my strong point ;-) )... thoughts:
- initial thought was `can we fix this by "'ting the $node`?
- but then I had a play, and found that we declare `node_basemen` with `declare -a`, whereas we are using it as an associative array I think, so should we be using `declare -A` instead? Maybe that will just make the string indexing work?, or maybe in combination with quoting the `$node` expansion, will make it work?

metrics/scaling/k8s_scale.sh
```diff
@@ -104,12 +104,17 @@ EOF
 		cpu_idle=${cpu_idle:-0}
 		mem_free=${mem_free:-0}
 
+		# bash arrays don't handle hyphens in the index as it treats
+		# a name like "azure-east-1" as 0-0-1, so remove any hyphens in the
+		# name, will work as long as hyphens aren't the only difference
+		node_index=$(tr -d [=-=] <<< $node)
```
grahamwhaley: I'd probably `s/-/_/` rather than remove them - that way the names are still readable when we debug etc.

dklyle:
I missed that the declaration was wrong. I changed that and life is good. I had tried every method of string escaping I could think of.

PR #165, opened 2019-08-29
metrics: scaling: require jq is installed

askervin:
Also remove trailing whitespace


metrics/scaling/k8s_scale.sh
```diff
@@ -200,7 +200,7 @@ EOF
 init() {
 	info "Initialising"
 
-	local cmds=("bc")
+	local cmds=("bc" "jq")
```
grahamwhaley: nice pickup.

PR #166, opened 2019-08-30
setup_system: add proxy configuration for docker

askervin:

### DIFF
--- a/clr-k8s-examples/setup_system.sh
+++ b/clr-k8s-examples/setup_system.sh
@@ -123,6 +123,24 @@ Environment="SOCKS_PROXY=${socks_proxy}"
 Environment="NO_PROXY=${no_proxy},${ADD_NO_PROXY}"
 EOF
 		done
+		# Add docker client proxy configuration for containers.
+		# If the configuration file already exists,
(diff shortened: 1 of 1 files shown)


krsna1729:
Was $RUNNER with docker tried? Also we don't install docker as part of this setup. Can you describe more in the commit and the first message in this conversation?

https://github.com/clearlinux/cloud-native-setup/blob/be164144e4c69c12cd2b2b4e835ec3cd21010839/clr-k8s-examples/setup_system.sh#L115-L124

askervin:
I'm running `metrics` scripts in a setup where clear-live-server has been just installed and `setup_system.sh` executed.

Running `cloud-native-setup/metrics/report/makereport.sh` needs docker. Launching `dockerd` will not be enough: successful run of `makereport.sh` needs docker client proxy configuration (or alternatively extra proxy handling software).

The point of this patch is to extend `setup_system.sh` so that `grabdata.sh` and `makereport.sh` run without any further setups or manual proxy configurations.

krsna1729:
move it into metrics. does not make sense to do it on every machine in the cluster.

askervin:
Thanks! Closing.

PR #173, opened 2019-09-03
adding detailed instructions for vagrant setup on clear linux

knangia:
Added command for installing unzip bundle since vagrant.sh requires unzip, else it will error out. Similarly, added command for installing rsync bundle, since Vagrantfile requires it. These bundles need to be added explicitly on some clear images. Also, removed the sudo, since running as normal user works fine.

### DIFF
--- a/clr-k8s-examples/vagrant.md
+++ b/clr-k8s-examples/vagrant.md
@@ -34,16 +34,15 @@ Note, vagrant installation steps were derived from:
 
 ## Install vagrant on Clear Linux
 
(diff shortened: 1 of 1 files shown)


CraigSterrett:
I would recommend starting with a sudo swupd update.
Then when installing using the clear vm image in the lab I found I was also missing wget and kvm-host.
Maybe instead of listing each separate you could have something like 
"Making sure prerequisite packages are installed"
swupd bundle-add unzip rsync wget kvm-host

knangia:
sounds good, will do.

CraigSterrett:
LGTM tested and the it all works now on lab VM

PR #189, opened 2019-09-23
Vagrant Libvirt Instructions update

ahsan518:
This patch adds the users to the default group so we dont
have to use sudo to start the vagrant, as using sudo is not
the most optimal route for gems to be properly configured.

Signed-off: Syed Ahsan <syed.ahsan.shamim.zaidi@intel.com>

### DIFF
--- a/clr-k8s-examples/vagrant.md
+++ b/clr-k8s-examples/vagrant.md
@@ -21,6 +21,10 @@ Download the latest Debian package from https://www.vagrantup.com/downloads.html
 sudo dpkg -i vagrant_${VER}_x86_64.deb
 vagrant plugin install vagrant-libvirt
 ```
+Make sure to add the user to the default group
+```bash
+usermod --append --groups libvirt `whoami`
+```
 
 Run vagrant
 ```bash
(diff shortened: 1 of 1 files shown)


AntonioMeireles:
humm... (hi all!)  


FWIW I already got that long ago covered by https://github.com/AntonioMeireles/ClearLinux-packer/blob/master/extras/clearlinux/setup/libvirtd.sh#L15-L17

at sight it's *just* the `sudo` calls which are redundant  and could be dropped

All the best, 

António

NitinAtIntel:
Per @AntonioMeireles , here's the PR https://github.com/AntonioMeireles/ClearLinux-packer/pull/20.
It pertains to a script used by instructions we've leveraged to run Vagrant/Libvirt on Clear Linux.  If we're not using that script for other host OS' then we may still need to ensure the user is added to the group, so that is one thing to check.
We can leave the removal of sudo I believe as is.

ahsan518:
Done

knangia:
Syed, we can still add the user in there and remove the sudo, since these instructions updated by you are for ubuntu, and not clear OS.

PR #193, opened 2019-09-25
Kustomize kubeadm.yaml config file

juliocri:
This change splits current configuration sections of kubeadm.yaml in single files in order to make it easier customizations using kustomize. create_stack script now build a new on the fly kubeadm.yaml file just before to execute kubeadm init.

### DIFF
--- a/clr-k8s-examples/create_stack.sh
+++ b/clr-k8s-examples/create_stack.sh
@@ -42,23 +42,19 @@ function finish() {
 trap finish EXIT
 
 function cluster_init() {
+	# Config replacements
 	if ! [ -z ${TOKEN} ]; then
-		sed -i "/InitConfiguration/a bootstrapTokens:\\n- token: ${TOKEN}" ./kubeadm.yaml
+		sed -i "/InitConfiguration/a bootstrapTokens:\\n- token: ${TOKEN}" ./kubeadm/base/kubeadm.yaml
 	fi
 	if ! [ -z ${MASTER_IP} ]; then
-		sed -i "/InitConfiguration/a localAPIEndpoint:\\n  advertiseAddress: ${MASTER_IP}" ./kubeadm.yaml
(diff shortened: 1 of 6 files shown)


jascott1:
Why don't we use Kustomize's patches instead of `sed`?

juliocri:
> Why don't we use Kustomize's patches instead of `sed`? I guess the env vars require it but we should consider patches for the other modifications.

Hey @jascott1 I think env vars can enter to the game of using patches without sed,  I submitted a commit with a proposal of implementation of patches using patchesJson6902 using yaml files as input.

mythi:
AFAICS, splitting `kubeadm.yaml` to three isn't necessary. I'd also suggest to use json style one-line patches that are all echoed in one patches .yaml.

jascott1:
I think we should just make a static kubeadm, keep the seds for env vars and use patches. Heredoc' ing everything is not a desirable way to generate yaml imo.

juliocri:
> I think we should just make a static kubeadm, keep the seds for env vars and use patches. Heredoc' ing everything is not a desirable way to generate yaml imo.

yeah, agree! I think this change is getting cleaner, `sed` are keep for env vars which injects its value, and patches for those other vars that are adding extra config such is the case of `$HIGH_POD_COUNT`

obedmr:
@juliocri, it seem like you will need to fix the conflicting files.

juliocri:
> @juliocri, it seem like you will need to fix the conflicting files.

Hey @obedmr Thanks for the follow up, what do you think is missing? maybe a commit squash? @jascott1

juliocri:
> yes, please squash

For sure! now this PR is squashed in a single commit.

juliocri:
@jascott1 ^_^

obedmr:
@jascott1 any comment on this? Is it ok for merge? it's blocking new requires changes for scaling efforts.

jascott1:
@obedmr review in progress :)

PR #196, opened 2019-09-26
Update Prometheus to v0.2.0

jascott1:


jascott1:
I tested this with k8s v1.15 and v1.16

hnanni:
tested with k8s 1.16 RUNNER as containerd. Worked fine, it is requiered to complete crete_stack.sh

PR #199, opened 2019-09-30
Update rook to v1.1.1

jascott1:


jascott1:
Tested on k8s v1.15 and v1.16

PR #200, opened 2019-10-01
metrics: add macros for sh debugging breakpoints

askervin:
This patch adds macros for setting breakpoints to shell scripts. Helps a lot when debugging local variable values.

Example:

1. start `nc -k -l -p 33720` to wait for the script to be debugged hitting a breakpoint

2. add `BP=my-breakpoint-name eval $BP_CALLHOME` anywhere where you wish the script to connect to the `nc` above, and give you interactive access to that point.


askervin:
No need to integrate to master, can included for development only when debugging is really needed.

Issue #202, opened 2019-10-03
some global variables are no longer defined


dklyle:
This manifests itself by seeing node_basemem[$node] array index issues.

PR #203, opened 2019-10-03
global variables in function aren't global

dklyle:
A recent change put global variable declarations in a function,
the variables no longer are globally visible. This change removes
the encapsulation in a function.

Closes: #202


obedmr:
LGTM, it's fixing our current CI issues. Will need to read more about this in-function variables definition behaviour.

PR #214, opened 2019-10-11
Adding support for aysnc metrics via collectd

dklyle:
This patch adds a new script, scaling/k8s_scaling_rapid.sh, for
launching pods and collecting metrics. The goal is to two fold.
The first goal is improve the required runtime duration of scaling
to large numbers of pods. k8s_scaling.sh can take up to 29 hours
to scale to 2900 pods. The is largely due to the overhead of
collecting system utilization stats after each new pod is launched.
This new script will collect system utilization stats asynchronously.
The second goal is to make it easier to collect additional system
utilization stats by leveraging the plugins supported by collectd.

Instead of using the stats daemon set, a new daemon set
that runs collectd on each node is added. collectd configuration
is handled by collectd/collectd.conf A configmap is added to the
K8s cluster containing collectd.conf, so the user of the script
can test new configurations easily. The configmap is created and
deleted as part of the test run. The data from collectd is stored on
each node in the cluster via the csv plugin and the data is
collected to the master node at the end of the test run.

Several new pages have been added to the metrics_report.pdf
These pages cover the same metrics as k8s_scale.sh results, but are
populated with data from collectd. Additionally, network interface
results are added.

To run the report, in addition to the previous steps of creating
a new directory and copying the result json file into it, all the
new <node_name>.tar.gz files must be copied in as well.

### DIFF
--- a/metrics/scaling/k8s_scale_rapid.sh
+++ b/metrics/scaling/k8s_scale_rapid.sh
@@ -1,0 +1,299 @@
+#!/bin/bash
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -e
+
+# Pull in some common, useful, items
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+source "${SCRIPT_PATH}/common.bash"
+source "${SCRIPT_PATH}/../collectd/collectd.bash"
+
+NUM_PODS=${NUM_PODS:-20}
+STEP=${STEP:-1}
+
+LABELVALUE=${LABELVALUE:-gandalf}
+
+pod_command="[\"tail\", \"-f\", \"/dev/null\"]"
+
+# Set some default metrics env vars
+TEST_ARGS="runtime=${RUNTIME}"
+TEST_NAME="k8s rapid"
+
+# $1 is the launch time in seconds this pod/container took to start up.
+# $2 is the number of pod/containers under test
+grab_stats() {
+	local launch_time_ms=$1
+	local n_pods=$2
+	local cpu_idle=()
+	local mem_free=()
+	local total_mem_used=0
+
+	info "And grab some stats"
+
+	local date_json="$(cat << EOF
+			"date": {
+				"ns": $(date +%s%N),
+				"Date": "$(date -u +"%Y-%m-%dT%T.%3N")"
+			}
+EOF
+	)"
+	metrics_json_add_array_fragment "$date_json"
+
+	local pods_json="$(cat << EOF
+			"n_pods": {
+				"Result": ${n_pods},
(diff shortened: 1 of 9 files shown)


obedmr:
Thanks @dklyle, I'm still reviewing/testing. Need to drop for Today but early on Monday will continue.

dklyle:
The network interface handling needs some work still as the interface naming changes across systems. I'll work on that Monday.

dklyle:
last commit includes a slight tweak to CPU collection and aggregation in collectd.conf

metrics/scaling/k8s_scale_rapid.sh
```diff
@@ -0,0 +1,299 @@
...
+	k8s_api_shutdown
+}
+
+show_vars()
+{
+	echo -e "\nEnvironment variables:"
+	echo -e "\tName (default)"
+	echo -e "\t\tDescription"
+	echo -e "\tTEST_NAME (${TEST_NAME})"
+	echo -e "\t\tCan be set to over-ride the default JSON results filename"
```
marcemq: `TEST_NAME` is not supported to over-ride the default results filename, from the code it uses a default value (`k8s rapid`), so we can add a note saying it or remove it entirely.

askervin:
I had some issues when running behind proxies. Collectd did not start because `apk update` tried to download packages via http without having `http_proxy` environment variable set inside the container.

What do you think, could we have something like:

*collectd.bash*
`kubectl create configmap collectd-proxy --from-literal http_proxy=$http_proxy --from-literal https_proxy=$https_proxy --from-literal no_proxy=$no_proxy`

*collectd.yaml*/spec/template/spec/containers/
```
          env:
            - name: http_proxy
              valueFrom:
                configMapKeyRef:
                  name: collectd-proxy
                  key: http_proxy
            - name: https_proxy
              valueFrom:
                configMapKeyRef:
                  name: collectd-proxy
                  key: https_proxy
            - name: no_proxy
              valueFrom:
                configMapKeyRef:
                  name: collectd-proxy
                  key: no_proxy
```

Also got problems when `/opt/collectd` did not exist on a node.

dklyle:
commit #5 changed the way we get the container image. Proxy issues should no longer exist.

msivosuo:
@obedmr, @askervin - ready to merge?

askervin:
Proxy issue is solved by using docker.io/dklyle/alpine-collectd image. I suppose it's ok as a workaround, but not a long term solution.

Test runs fine on two-node (master-worker1) cluster and gathers master.tar.gz and worker1.tar.gz.

Report script has an issue. Runtime scaling rapid page in the PDF gets:
```
## Warning in file(file, "rt"): cannot open file '/inputdir/rapid/worker2/
## localhost/memory/NA': No such file or directory
```

(worker2 node was not used in the test run, but there's a NotReady worker2 node visible in `kubectl get nodes`)

I'm ok with merging this patch and addressing these issues in later fixes.

Issue #221, opened 2019-10-15
json: unknown field "metadata" 


CraigSterrett:
It appears this was introduced with https://github.com/clearlinux/cloud-native-setup/pull/193 

When running create_stack.sh init I get the following
W1015 23:25:59.562402    5109 strict.go:54] error unmarshaling configuration schema.GroupVersionKind{Group:"kubeadm.k8s.io", Version:"v1beta1", Kind:"InitConfiguration"}: error unmarshaling JSON: while decoding JSON: json: unknown field "metadata"
W1015 23:25:59.563404    5109 strict.go:54] error unmarshaling configuration schema.GroupVersionKind{Group:"kubelet.config.k8s.io", Version:"v1beta1", Kind:"KubeletConfiguration"}: error unmarshaling JSON: while decoding JSON: json: unknown field "metadata"
W1015 23:25:59.564377    5109 strict.go:54] error unmarshaling configuration schema.GroupVersionKind{Group:"kubeadm.k8s.io", Version:"v1beta1", Kind:"ClusterConfiguration"}: error unmarshaling JSON: while decoding JSON: json: unknown field "metadata"

I have reproduced this using both Clear version 31130 (K8s 1.15) and 31300 (K8s 1.16)

knangia:
Just edited it and inserted code ^^

ahsan518:
Just verified this, and got the same error.

knangia:
I get the same error when running create_stack.sh on master node with Clear version: 31310 and k8s v1.16.2 using the vagrant setup as mentioned in the main repo.
```
W1015 23:59:14.123150    3962 strict.go:54] error unmarshaling configuration schema.GroupVersionKind{Group:"kubeadm.k8s.io", Version:"v1beta1", Kind:"ClusterConfiguration"}: error unmarshaling JSON: while decoding JSON: json: unknown field "metadata"
W1015 23:59:14.125888    3962 strict.go:54] error unmarshaling configuration schema.GroupVersionKind{Group:"kubeadm.k8s.io", Version:"v1beta1", Kind:"InitConfiguration"}: error unmarshaling JSON: while decoding JSON: json: unknown field "metadata"
W1015 23:59:14.129216    3962 strict.go:54] error unmarshaling configuration schema.GroupVersionKind{Group:"kubelet.config.k8s.io", Version:"v1beta1", Kind:"KubeletConfiguration"}: error unmarshaling JSON: while decoding JSON: json: unknown field "metadata"
[...]

PR #222, opened 2019-10-16
Fixes unmarshal errors from kubeadm init with kustomize

juliocri:
This fixes: #221

Removing error logs from `kubeam init` caused from mismatched json fields.


juliocri:
Seems that with this flag `--experimental-kustomize` the kustomization dir is taken, but patches over controller manager are not reflected in the final config.

jascott1:
@juliocri did you get it the kustomize to work w the corrected path? If so, what about something like:

```
KUBEADM_CONFIG="base"
if [[ -n $HIGH_POD_COUNT ]]; then
  KUBEADM_CONFIG="high-pod-count"
fi
sudo -E kubeadm init --experimental-kustomize=./kubeadm/"$KUBEADM_CONFIG"
```

juliocri:
> @juliocri did you get it the kustomize to work w the corrected path? If so, what about something like:
> 
> ```
> KUBEADM_CONFIG="base"
> if [[ -n $HIGH_POD_COUNT ]]; then
>   KUBEADM_CONFIG="high-pod-count"
> fi
> sudo -E kubeadm init --experimental-kustomize=./kubeadm/"$KUBEADM_CONFIG"
> ```

@jascott1 no, it does not work, `--experimental-kustomize` by now only patches static pods manifests [here the doc,](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cluster-lifecycle/kubeadm/20190722-Advanced-configurations-with-kubeadm-(Kustomize).md#kustomize-integration-with-kubeadm) it does not apply patches on other yamls such the kubeadm.yaml for example...

jascott1:
Maybe `sed` out the names after templating as you suggested or just go back to `sed` altogether. :|

clr-k8s-examples/create_stack.sh
```diff
@@ -63,6 +63,7 @@ function cluster_init() {
 		echo "/var/lib/etcd exists! skipping init."
 		return
 	fi
+	echo "$(kubectl kustomize ./kubeadm/metadata/)" > ./kubeadm/base/kubeadm.yaml
```
juliocri: @jascott1 a last try patching, how do this looks like? or we should move back to `sed`? as you suggest in your last comment, and let this kubeadm-kustomize gets more mature.

juliocri:
@jascott1 this final commit removes kustomize support from kubeadm config and comes back to the `sed` replacement method. can you take a look in a chance? :)

PR #223, opened 2019-10-16
Add extra table about node information into PDF report.

marcemq:
**What**
A final table report gets added into the PDF report for node information details per node.

**Tests**
The changes where tested using two different results, following the next structure:
```
$ tree metrics/results/
metrics/results/
├── Env.R
├── scaling
│   └── k8s-scaling.json
└── scaling-vms
    └── k8s-scaling.json
``` 
The final look for the created tables is shown below 🎉 
node-1
And the final PDF report is attached too.
[metrics_report.pdf](https://github.com/clearlinux/cloud-native-setup/files/3736493/metrics_report.pdf)

### DIFF
--- a/metrics/report/report_dockerfile/metrics_report.Rmd
+++ b/metrics/report/report_dockerfile/metrics_report.Rmd
@@ -47,3 +47,13 @@ in the test results files.
 ```{r dut, echo=FALSE, fig.cap="System configuration details"}
 source('dut-details.R')
 ```
+
+\pagebreak
+
+# Test setup node details
+
+This table describes node details within the Kubernetes cluster that have been used for test.
+
+```{r node, echo=FALSE, fig.cap="Node information within Kubernetes cluster"}
+source('node-info.R')
+```
--- a/metrics/report/report_dockerfile/node-info.R
+++ b/metrics/report/report_dockerfile/node-info.R
@@ -1,0 +1,95 @@
+#!/usr/bin/env Rscript
+# Copyright (c) 2019 Intel Corporation
+#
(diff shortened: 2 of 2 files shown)


grahamwhaley:
silly question - I've not looked at the code - what happens if two runs/stacks happen to have nodes that have the same name - will they appear as different entries in the table?
I can conceive this happening at my desk for instance, where I might run up a stack, do a test, reconfig the same stack, and do the test again... same stack/nodes, but with different setups ;-)

marcemq:
There aren't silly questions my dear @grahamwhaley 😃 

To your question, and having the same node with different setups, i.e.:

- `NUM_PODS=10 grace=0 ./scaling/k8s_scale.sh`
- `NUM_PODS=20 grace=0 ./scaling/k8s_scale.sh`

And the `results` directory have the next distribution:
```
$ tree results/
results/
├── Env.R
├── scaling-P10
│   └── k8s-scaling.json
└── scaling-P20
    └── k8s-scaling.json

2 directories, 3 files
```
And the resulting report shows different entries for the node details, even that the info is the same but the tests are different, which is the same por system config details, see table below:
node-1
The generated report is also attached 
[metrics_report.pdf](https://github.com/clearlinux/cloud-native-setup/files/3752016/metrics_report.pdf)

From the generated report we can noticed that:

- there is a bug for System CPU percentage, we cannot report negative values for those right?
- there is a bug for labels shown at Pod boot time, I think it needs to be the same as System CPU usage (`testname` and `node`) right?

I'll open issues for the ones mentioned above.

Issue #230, opened 2019-10-21
makereport.sh reports negative entries for CPU usage


marcemq:
See generated report attached for System CPU usage graph.
[metrics_report.pdf](https://github.com/clearlinux/cloud-native-setup/files/3752869/metrics_report.pdf)

dklyle:
The problem here is sample size. But the negative value is correct in this case.

PR #231, opened 2019-10-21
Correct compute of CPU usage.

marcemq:
**What**
Fix for negative entries reported at System CPU usage graph (see #230 ), which takes into account the absolute difference between first and last entry for `idle` data.

**Testing**
Having the next result entries:
```
$ tree results/
results/
├── Env.R
├── scaling-P10
│   └── k8s-scaling.json
└── scaling-P20
    └── k8s-scaling.json

2 directories, 3 files
```

The report execution generates the next graph for System CPI usage:
scaling-2
As it can be see, the scaling execution can start with a higher CPU usage, maybe for previous tests or it's is simply the current state for the nodes, or viceversa, that's why to consider the absolute difference instead of a simple subtraction operation. 
The complete generated report is attached too ([metrics_report.pdf](https://github.com/clearlinux/cloud-native-setup/files/3752985/metrics_report.pdf))

### DIFF
--- a/metrics/report/report_dockerfile/tidy_scaling.R
+++ b/metrics/report/report_dockerfile/tidy_scaling.R
@@ -140,7 +140,7 @@ for (currentdir in resultdirs) {
 					next
 				}
 				memtotal = memtotal + thisnode[nrow(thisnode),]$mem_used
-				cpuused = thisnode[1,]$idle - thisnode[nrow(thisnode),]$idle
+				cpuused = thisnode[nrow(thisnode),]$idle - thisnode[1,]$idle
 				cputotal = cputotal + cpuused
 				inodetotal = inodetotal + thisnode[nrow(thisnode),]$inode_used
 			}


dklyle:
The absolute value doesn't make sense here. If we actually free up 4.5% of the CPU when the test runs, then the value is -4.5%. We aren't consuming 4.5%. Another option is to take the max(CPU) - min(CPU). The problem here is the data set size. 20 pods is not going to generate enough load to influence the CPU utilization significantly. These tests also look to be run on an all-in-one cluster. The period sync operations done on the K8s master node can account for all the CPU swing here. 

My thought is, while not the result we may expect, the result is correct. We don't alter the data or report to show what we expect, we want it to show what actually happened.

grahamwhaley:
I concur with @dklyle - the core problem is that the test itself and the test run size are not large/hard enough to make much of an impact on the CPU utilisation. I think showing the 'real numbers' as is is the best thing we can do.

dklyle:
The metric, thisnode$idle is the % of time the CPU is in the idle state. The utilization number in the end is obtained by taking 100% - the idle%. What you are proposing would be correct if the metric was thisnode$utilized, but since we are tracking idle, the order is currently correct to subtract start_idle - end_idle as CPU idle should decrease for any significant number of pods.

marcemq:
Huge thanks for the clarification @dklyle!! 😊 

I'm proceeding with closing this PR.

PR #233, opened 2019-10-22
disregard collectd tail data from stats

dklyle:
Since collectd is started before the pods are launched and
shutdown after the last pod is launched, we gather data outside
the pod launch window which can adversely influence the per pod
launch stats. This is especially true after the last pod launches
as all the pods are then deleted before collectd stops collecting
metrics.

This patch isolates the collectd data used to only
coincide with the pod launch window.

An additional change in this patch is to improve the secondary
y axis scaling. There was an ill-advised check in previously to
force the scale to be at least 1. This does not work well when
the pod number is significantly higher than say 100 (the max
possible cpu idle value).

This patch changes the scaling to be across all data to be graphed.
The special condition remains for interface drops and interface errors,
where the data is typically 0. So we don't scale by 0.

### DIFF
--- a/metrics/report/report_dockerfile/collectd_scaling.R
+++ b/metrics/report/report_dockerfile/collectd_scaling.R
@@ -30,15 +30,6 @@ cpustats=c()		# Statistics for cpu usage
 bootstats=c()		# Statistics for boot (launch) times
 inodestats=c()		# Statistics for inode usage
 
-# values for scaling the secondary y axes on some graphs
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@askervin @marcemq - I'd like to land this, as I've run into the 'badly scaled double y axis' issue twice, that is fixed in this PR.
Can you have a quick review before I land it. thx!

askervin:
lgtm

PR #239, opened 2019-10-24
Adding Flannel to CNI

ahsan518:
This patch add's Flannel to CNI, export the environment variable
to use Flannel, and uses crio as default unless the user
specify otherwise.

Signed-Off: Syed Ahsan <syed.ahsan.shamim.zaidi@intel.com>

#190


jascott1:
LGTM Manually tested on k8s 1.16.2

PR #252, opened 2019-10-28
Specify hostname for file name for collectd

dklyle:
On some machines the value for hostname, that is used in naming
the csv directory, isn't always determined by collectd to be
localhost. The scaling code assumes it will always be localhost.

This patch specifies the hostname to be localhost.

### DIFF
--- a/metrics/collectd/collectd.conf
+++ b/metrics/collectd/collectd.conf
@@ -9,6 +9,8 @@ LoadPlugin memory
 LoadPlugin cpufreq
 LoadPlugin df
 
+Hostname localhost
+
 <Plugin "cpu">
     ReportByCpu true
     ReportByState true


dklyle:
This needs more verification on a system where localhost is not the value by default. Won't merge until completed.

grahamwhaley:
I gave this a spin. It now generates data and a report pdf, but I see the following at the start of the 'rapid' section in the report:

```
## Warning in max(interface_packets_data$tx, na.rm = TRUE): no non-missing
## arguments to max; returning -Inf
```

repeated a number of times. It could be unrelated...

For reference then, the issues I've been seeing are because I've been testing with a [kind](https://github.com/kubernetes-sigs/kind) cluster. Here are some details of my setup:

My `kind-config.yaml` looks like:

```yaml
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
nodes:
# the control plan
        - role: control-plane
# the worker
        - role: worker
```

In this instance I have one worker and one master, but I do also run with multiple workers (normally 2).

I kick off the cluster using:

```bash
$ kind create cluster --config $(pwd)/kind-config.yaml
```

and set my env with

```bash
$ export KUBECONFIG="$(kind get kubeconfig-path)"
```

and then run the scaling scripts as normal.

dklyle:
I tested without this PR on a KIND setup and it is not causing the above error. See #257

grahamwhaley:
@dklyle - I think you and I figured out the `-inf` stuff was not related, didn't we.
@obedmr @marcemq @askervin - ptal - nice simple fix :-)

dklyle:
@grahamwhaley yes, the warning was unrelated to this particular issue.

obedmr:
LGTM, tested and worked as expected.

obedmr:
@grahamwhaley, @marcemq, @askervin  is it ok for you to merge?

askervin:
Tested. Looks good and works fine.

PR #256, opened 2019-10-29
fixing makereport container build

dklyle:
A recent change to debian apt repositories led to build errors
for the report container. The error was around stretch release
files. The container image we are based on published an update
which fixes this error. This patch updates to use :latest to
avoid errors when building.


grahamwhaley:
@askervin @marcemq WDTY .... as generally I'd like more than one ACK on each PR ;-)

marcemq:
@grahamwhaley what does it mean WDTY? and ACK?

grahamwhaley:
s/WDTY/WDYT/ -> 'What Do You Think' :-)
ACK -> Ack(nowledgement) - as in ack/nack in transmission protocols etc. :-) https://www.linktionary.com/n/nak.html

marcemq:
LGTM 😊

PR #258, opened 2019-10-29
support older network interface naming

### LINKED ISSUE #257: KIND uses old network interface naming pattern
The default KIND image still uses the 'eth' prefix for the network interconnect. The result is no network interface data is recorded. The error when creating the report looks like:

## arguments to max; returning -Inf

dklyle:
KIND uses the older network interface naming standard. Other
operating system images may as well. Adding support for 'eth'
network interface naming prefix.

Closes #257 


grahamwhaley:
pppping @askervin @marcemq ..... ;-)

PR #265, opened 2019-11-11
Fixed code for no_proxy in setup_system.sh

CraigSterrett:
The sed command for updating no-proxy settings in the proxy.sh file was
not working. I made it match the above command which was working for
/etc/environment.

Signed-off-by: Craig Sterrett <craig.sterrett@intel.com>


CraigSterrett:
Trying to locate another test environment. Testing on my NUC the code works, but the stack rarely runs to completion due to lack of resources (Master VM usually crashes). If I get access to another test environment I can share that so somebody else can test my code change.

CraigSterrett:
I built up a test system with sufficient resources and Knangia is testing now

PR #269, opened 2019-11-13
set the snapshotter to devmapper in setup script

egernst:
Signed-off-by: Eric Ernst <eric.ernst@intel.com>

### DIFF
--- a/clr-k8s-examples/containerd_devmapper_setup.sh
+++ b/clr-k8s-examples/containerd_devmapper_setup.sh
@@ -56,6 +56,9 @@ else
   [plugins.devmapper]
     pool_name = "contd-thin-pool"
     base_image_size = "4096MB"
+  [plugins.cri]
+    [plugins.cri.containerd]
+      snapshotter = "devmapper"
 EOT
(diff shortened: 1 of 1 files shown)


clr-k8s-examples/containerd_devmapper_setup.sh
```diff
@@ -56,7 +56,10 @@ else
   [plugins.devmapper]
     pool_name = "contd-thin-pool"
     base_image_size = "4096MB"
-EOT
+  [plugins.cri]
+    [plugins.cri.containerd]
+      snapshotter = "devmapper" 
+    EOT
```
grahamwhaley: EOT got indented - not sure if that is actually a real problem though.

egernst:
@grahamwhaley if the user is configuring containerd's snapshotter (ie, calling this script), i'm assuming they want to actually use the snapshotter.

Without this, containerd continues to leverage overlay, ignoring the thinpool created and disk images provided...

egernst:
@ganeshmaharaj PTAL?

Issue #272, opened 2019-11-15
Confusion with change to CLRK8S_RUNNER in create_stack.sh


CraigSterrett:
create_stack.sh uses CLRK8S_RUNNER where setup_system uses RUNNER. This needs to be unified and the README updated to reflect the unification.

PR #273, opened 2019-11-18
Using consistent ENV variable for RUNNER

ahsan518:
Earlier there was confusion between CLRK8S_RUNNER and RUNNER in
setup_system.sh and create_stack.sh script, this patch fixes this
and now the variable can either be RUNNER/CLRK8S_RUNNER.

Signed-off: syed.ahsan.shamim.zaidi@intel.com 
#272

### DIFF
--- a/clr-k8s-examples/README.md
+++ b/clr-k8s-examples/README.md
@@ -38,6 +38,9 @@ This script ensures the following
 script uses the runtime specified in the `RUNNER` environment variable and defaults to `crio`. To use the
 `containerd` runtime, set the `RUNNER` environment variable to `containerd`.
 
+In case of vagrant, if you want to spin up VM's using different environment variable than declared in [`setup_system.sh`],
+specify when performing vagrant up. E.g., `RUNNER=containerd vagrant up`
+
 ### Specify a version of Clear Linux
 
 To specify a particular version of Clear Linux to use, set the CLRK8S_CLR_VER environment variable to the desired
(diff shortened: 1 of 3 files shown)


clr-k8s-examples/README.md
```diff
@@ -38,6 +38,9 @@ This script ensures the following
 script uses the runtime specified in the `RUNNER` environment variable and defaults to `crio`. To use the
 `containerd` runtime, set the `RUNNER` environment variable to `containerd`.
 
+In case of vagrant, if you want to spin up VM's using different environment variable then declared in [`setup_syste.sh`],
```
grahamwhaley: s/the/than/
s/setup_syste.sh/setup_system.sh/

clr-k8s-examples/create_stack.sh
```diff
@@ -30,7 +30,9 @@ METALLB_VER="${CLRK8S_METALLB_VER:-v0.8.1}"
 NPD_VER="${CLRK8S_NPD_VER:-v0.6.6}"
 PROMETHEUS_VER="${CLRK8S_PROMETHEUS_VER:-f458e85e5d7675f7bc253072e1b4c8892b51af0f}"
 CNI=${CLRK8S_CNI:-"canal"}
+if [[ ! -z "$RUNNER" ]]; then RUNNER="${RUNNER}"; fi
```
grahamwhaley: This looks benign? I read it as 'if runner is non zero set runner to runner. Am I missing some subtlety?

clr-k8s-examples/create_stack.sh
```diff
@@ -144,10 +146,10 @@ function cni() {
 		FLANNEL_VER=${1:-$FLANNEL_VER}
 		FLANNEL_URL="https://github.com/coreos/flannel"
 		FLANNEL_DIR="0-flannel"
-		
+
```
grahamwhaley: suggestion: maybe put whitespace fixes in a separate commit in the future - it makes reviewing the difference between a whitespace fix and a technical fix easier ;-)

clr-k8s-examples/create_stack.sh
ahsan518: This is used in case for Cilium, since cilium is using Runner env variable, so if runner is not declared, it uses the Runner as set by setup_system.sh

clr-k8s-examples/create_stack.sh
ahsan518: I usually after doing changes do a shfmt check.

PR #274, opened 2019-11-19
scaling: Remove tty parameter in report's generation cmd

obedmr:
This is in order to avoid tty-related issues in our CI systems
which by default is not supporting tty. With this change we'll
avoid the following faling report's generation `docker run` command.
```
the input device is not a TTY
```

Signed-off-by: Obed N Munoz <obed.n.munoz@intel.com>

### DIFF
--- a/metrics/report/makereport.sh
+++ b/metrics/report/makereport.sh
@@ -100,7 +100,7 @@ setup() {
 }
 
 run() {
-	docker run -ti --rm -v ${HOSTINPUTDIR}:${GUESTINPUTDIR} -v ${HOSTOUTPUTDIR}:${GUESTOUTPUTDIR} ${extra_volumes} ${IMAGE} ${extra_command}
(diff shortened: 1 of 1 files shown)


obedmr:
@grahamwhaley , @askervin, @marcemq, @dklyle, if you have a chance, would you mind having a look on this?

askervin:
I think this is pretty good. One suggestion to keep the interactive debug mode working, though:

How about having something like:
```
docker run ${extra_opts} --rm -v ${HOSTINPUTDIR}:${GUESTINPUTDIR} -v ...
           ^^^^^^^^^^^^^
```
and then in the debug mode (option `-d`) after `extra_volumes=...`
```
                        extra_opts="-ti"
```

grahamwhaley:
ack - @obedmr good catch - I've hit that before with the CIs and the kata tests, and had to purge all the tty's from there :-). I agree with @askervin though, that we still need the tty for the interactive debug mode - pls add that into the commit :-)

obedmr:
@grahamwhaley @askervin thanks for the feedback. It's implemented and ready for the second shot.

grahamwhaley:
Tested using `nohup ./makereport.sh` :-) Merging.

PR #275, opened 2019-11-20
metrics: make pods_per_gb valid JSON

grahamwhaley:
Under some circumstances, the pod_per_gb value would come out as <0,
and be generated without any leading 0's (such as `.14` rather than
`0.14`). This is not valid JSON, and would break the report generation
parsing in R.

Use `printf` to force a leading 0 prefix onto the value.

### DIFF
--- a/metrics/scaling/k8s_scale.sh
+++ b/metrics/scaling/k8s_scale.sh
@@ -169,7 +169,7 @@ EOF
 	if [ $n_pods -eq 0 ]; then
 		local pods_per_gb=0
 	else
-		local pods_per_gb=$(bc -l <<< "scale=2; ($total_mem_used/1024) / $n_pods")
+		local pods_per_gb=$(printf "%0f" $(bc -l <<< "scale=2; ($total_mem_used/1024) / $n_pods"))
 	fi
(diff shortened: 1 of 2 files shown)


grahamwhaley:
Spotted during the auto CI pdf generation from @obedmr 
Example: if you run:

```bash
#!/bin/bash

set -x

total_mem_used=1024
n_pods=7

test() {
        local pods_per_gb=$(bc -l <<< "scale=2; ($total_mem_used/1024) / $n_pods")
        local new_pods_per_gb=$(printf "%0f" $(bc -l <<< "scale=2; ($total_mem_used/1024) / $n_pods"))
}

test
```

you get:

```bash
+ total_mem_used=1024
+ n_pods=7
+ test
++ bc -l
+ local pods_per_gb=.14
+++ bc -l
++ printf %0f .14
+ local new_pods_per_gb=0.140000
```

Here is some JSON to show the error and the correction:

```json
{
    "bad": .14,
    "good": 0.140000
}
```

To aid any future searches, the error was similar to `Error in parse_con .... invalid char in json text`

askervin:
Works well, LGTM.

PR #276, opened 2019-11-21
metrics: scaling: fix R when STEP>1 && NODES>1

grahamwhaley:
Recently we'd seen a few report generation errors, particularly it seemed when we had STEP>1 and NODES>1. The tidy_scaling.R code was wrong when trying to handle node names and pod launches when we had more than one pod launched (STEP>1), and they launched on multiple different nodes (NODES>1).
Fix by re-coding.

There was also an `df` inodes but that prevented the `k8s_scaling.sh` running on some clusters (in my case, a 3-node kind cluster).
Fixed by 'pinning' the inode collection to just the root mount.

### DIFF
--- a/metrics/report/report_dockerfile/tidy_scaling.R
+++ b/metrics/report/report_dockerfile/tidy_scaling.R
@@ -66,8 +66,7 @@ for (currentdir in resultdirs) {
 			########################################################
 			local_bootdata=tibble(launch_time=br$launch_time$Result)
 			local_bootdata=cbind(local_bootdata, n_pods=br$n_pods$Result)
-			local_bootdata=cbind(local_bootdata, node=lp$node)
-			local_bootdata=cbind(local_bootdata, testname=rep(testname, length(local_bootdata$node)))
+			local_bootdata=cbind(local_bootdata, testname=rep(testname, length(local_bootdata$n_pods)))
 
 
 			########################################################
@@ -115,7 +114,7 @@ for (currentdir in resultdirs) {
 			local_nodedata=cbind(local_nodedata, mem_used=used_df$Result)
 			local_nodedata=cbind(local_nodedata, inode_free=ifree_df$Result)
 			local_nodedata=cbind(local_nodedata, inode_used=iused_df$Result)
-			local_nodedata=cbind(local_nodedata, testname=rep(testname, length(local_nodedata$node)))
+			local_nodedata=cbind(local_nodedata, testname=rep(testname, length(local_nodedata$n_pods)))
 
 			# Now Calculate some stats. This gets more complicated as we may have n-nodes,
 			# and we want to show a 'pod average', so we try to assess for all nodes. If
@@ -124,15 +123,15 @@ for (currentdir in resultdirs) {
 			# table.
 
 			# Get a list of all the nodes
-			nodes=unique(local_nodedata$node)
+			unique_nodes=unique(local_nodedata$node)
 
 			memtotal=0
 			cputotal=0
 			inodetotal=0
 			# Calculate per-node totals, and tot them up to a global total.
-			for (n in nodes) {
+			for (n in unique_nodes) {
 				# Make a frame with just that nodes data in
-				thisnode=subset(local_nodedata, node %in% c(n))
+				thisnode=local_nodedata[ which(local_nodedata$node==n),]
 
 				# Do not use the master (non-schedulable) nodes to calculate
 				# launched pod metrics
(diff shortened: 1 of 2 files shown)


grahamwhaley:
@dklyle - you might want to cast a glance over the stats inode fix. It's not perfect, but improves the situation as is I think. Input welcome.

marcemq:
**Testing done**

Having a all-in-one Kubernetes clusted:
* cloned latest `grahamwhaley/cloud-native-setup` 
* checkout to `20191120_bad_arrays` branch with commit `b5e946ddf03be0a`
* execute scale and rapid tests with default variables and `STEP` variable (`STEP=2./scaling/k8s_scale.sh` and `STEP=2 ./scaling/k8s_scale_rapid.sh` )
* created the required results directory using **all** above output results which looks like:
```
results/
├── Env.R
└── test-PR276
    ├── clr-30f01b5149ba4ab8b05a7ee03b6812a5
    ├── k8s-rapid.json
    └── k8s-scaling.json
```

* hit `./report/makereport.sh` to create the report.

**Findings**:

* PDF gets created for scaling and rapid tests ✅ 
* The test setup node details does not get generated ❌ 

The generated report is attached below:
[metrics_report.pdf](https://github.com/clearlinux/cloud-native-setup/files/3917636/metrics_report.pdf)

@grahamwhaley what are the exact steps to test your changes? should I need to setup a Kubernetes cluster with more nodes for your changes? 
And for further changes, please try to have separate PRs for separate fixes so it's easier to test them out.

grahamwhaley:
Hi @marcemq . Thanks for the feedback.
I just tested against the HEAD code, commit `bdcb4fb (HEAD -> master, origin/master, origin/HEAD) Update Kata to v1.9.1 (#282)`, and the same `## Error in fdata$socketsPerNode[i, ]: incorrect number of dimensions` error exists for me in the node details as it does in your PDF.

The fundamental reason would seem to be that [this code](https://github.com/clearlinux/cloud-native-setup/blob/master/metrics/report/report_dockerfile/node-info.R#L48)

```R
for (i in 1:numnodes) {
  node_i=fdata$'kubectl-get-nodes'$items[i,]
  node_info=fdata$'socketsPerNode'[i,]
  ...
```

is trying to read from an array, whereas in the actual `k8s-scaling.json`, we find:

```json
    "kind": "List",
    "metadata": {
        "resourceVersion": "",
        "selfLink": ""
    }
},
        "date" : {
                "ns": 1575391781499192201,
                "Date": "2019-12-03T16:49:41.500"
        },
        "test" : {
                "runtime": "default",
                "testname": "k8s rapid"
        },
        "socketsPerNode": [

        ],
"k8s-rapid" : {
...
```

So, it looks like at least under some circumstances (that both you and I currently have with our clusters), that `socketsPerNode` array is empty, and thus I think the 'indexing of node1' in the `R` code bombs.

afaik, that is not related to this PR and patches - it exists in the HEAD of the codebase.

So, sure, I can/could break all not-quite-related commits into separate PRs, if we think that will actually help. What will also help is, when testing, you could:
- first create the problem that the PR is saying it fixes, and then try it with the PR to see if it really did fix it
- if you find an issue with a multiple-commit PR then you could
  - check that the issue does not exist in the HEAD of the codebase (e.g., potentially unrelated to the PR)
  - [bisect](https://git-scm.com/docs/git-bisect) the commits to identify which one is causing the issue
    - and then do the first stage diagnosis to see if you can feed some direction back to the author (and check that you understand what the commit is doing, just in case it is an obvious misunderstanding).

Of those, right now afaict, this PR does not make your report any worse for your cluster and json results report generation. If you want to perform a more thorough test, then as the commit message says, this PR fixes a situation where we had NODES>1 and STEPS>1 - so you'll need to make a cluster with >1 node, and run the tests with STEP>1 - see that you can witness the issue with the `HEAD` code, and then check if this PR fixes that.....

marcemq:
I agree that the shown error at the PDF report is not related with this PR. I'll setup a Kubernetes cluster with `NODES>1` and shown my findings.

And yes @grahamwhaley having separate PRs for nor-quite-related commits are helpful. And ideally, we should have an open bug that has the spotted error and the **exact steps to reproduce the bug**  and then, when sending a fix, we can simply refer to that bug in order to test that such bug gets fixed. In the case when not having a bug open, having the exact steps to test the changes is really helpful, it will help our velocity when reviewing the PRs and it will serve as a clear and good guide for new developers who wants to review a PR which is great for open source projects 😊

grahamwhaley:
Closing. work on hold.

PR #277, opened 2019-11-22
[RFC | WIP] metrics: Experimental jupyter plotly interactive rendering

grahamwhaley:
Set up a Jupyter R based docker image, with some additions, and use
that to render the ggplot output of the R(markdown) metrics scripts
as interactive ploty views.

### DIFF
--- a/metrics/jupyter/README.md
+++ b/metrics/jupyter/README.md
@@ -1,0 +1,50 @@
+# Jupyter image and playbooks
+
+This folder contains tools that enable the metrics report R
+ggplot graphs to be loaded into a Jupyter playbook and examined as `plotly`
+images. This allows elements (data sets) be be interactively en/disabled in the
+graph, as well as selection and zooming of the view. This can aid analysis of
+complex and noisy graphs.
+
+# Running the Jupyter
+
+You require `Docker` to run the playbooks. The folder containers a `Dockerfile`,
+and will build/install the required Jupyter image, along with the required R libaries
+needed to process and display the report graphs.
+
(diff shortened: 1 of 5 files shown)


grahamwhaley:
@dklyle @askervin @obedmr @marcemq (/cc @msivosuo)

Something I've been playing with in the background - given some of our graphs can now get quite dense and complex with long runs and many data sets, I wondered if we could get an 'interactive data view' of them so we could turn on/off some data sets and zoom in on interesting bits of the data...

To that end, I had a play with Jupyter notebooks, and found we can do a simple first pass by wrapping our ggplot output into a (gg)plotly wrapper. That gives us enablable data items (turn items on/off) and zoomability - and also the ability to save out a png file (like below, a zoom of our recent CI 2k pod run).

It's not perfect, hence WIP/RFC - but, maybe you can find it useful, and maybe we can improve it over time.
I will continue to tickle the idea of using other packages (shiny, rbokeh etc.) to see if we can do a better job - but they seem much more complex to enable much more functionality. For instance, for some graphs I'd really like a 'smoothed line' where we can adjust the smoothing factor on a slider - I think this could really help evaluate any trends and bumps in the boot graphs for instance - but, that is non-trivial to write with our R based code right now.

rapid_cpu_zoomed

grahamwhaley:
Oh, I should note (in the README), that I had to disable selinux before I could mount the volumes into the docker image - a pain, but there you go... I didn't have time to figure out single exceptions to add to selinux

grahamwhaley:
Closing, work on hold.

PR #212, opened 2019-10-11
metrics: report pod communication latency percentiles in pdf

askervin:
- Add R routine to load and plot percentiles from JSON.
- Support any number of percentiles 1..n.
- Store percentile configuration in JSON.
- Fix handling percentile value 100.

Signed-off-by: Antti Kervinen <antti.kervinen@intel.com>

PR #278, opened 2019-11-22
Update test description for rapid and nc scaling tests.

marcemq:


metrics/scaling/k8s_scale_nc.sh
```diff
@@ -437,8 +437,8 @@ help()
 	usage=$(cat << EOF
 Usage: $0 [-h] [options]
    Description:
-	Launch a series of workloads and take memory metric measurements after
-	each launch.
+	Launch a series of workloads and take pod query-response latency measurements
```
juliocri: IMHO `query-response` can be redundant, latency in most of the cases refers to the time that something takes to response.

metrics/scaling/k8s_scale_rapid.sh
```diff
@@ -211,8 +211,8 @@ help()
 	usage=$(cat << EOF
 Usage: $0 [-h] [options]
    Description:
-	Launch a series of workloads and take memory metric measurements after
-	each launch.
+	Launch a series of workloads and take memory metric measurements using collectd
```
juliocri: a little suggestion: using single quotes to refer another command/application e.g. `collectd` to avoid confusion with a typo.

metrics/scaling/k8s_scale_nc.sh
marcemq: I've taken this description from the ongoing [PR212](https://github.com/clearlinux/cloud-native-setup/pull/212/files#diff-49c99b2e785db0c2676430f5704887deR54) of such test. 

@askervin any thoughts on this?

metrics/scaling/k8s_scale_rapid.sh
marcemq: Done.

askervin:
LGTM.

PR #212 changes the network latency test script name from `test_scale_nc.sh` to `test_rapid_nc.sh`, as it is based on the rapid/collectd test now - instead of the original `grab_stats`. However, having this change merged will remind me (in the form of a merge conflict) that PR #212 should change the script documentation as well.

grahamwhaley:
> _No description provided._

Please provide meaninful descriptions in the PR and the commit bodies, even if they are brief and mostly are repeat of the subject - having a 'what and why and how' in the git logs helps in the future.

metrics/scaling/k8s_scale_nc.sh
grahamwhaley: I think `query-response` is fine (although you may wish to add 'http' or whatever protocol is being used to it). Without saying `query-response`, then I don't think we'd know which latency was being measured (network, boot, etc.).

metrics/scaling/k8s_scale_rapid.sh
```diff
@@ -211,8 +211,8 @@ help()
 	usage=$(cat << EOF
 Usage: $0 [-h] [options]
    Description:
-	Launch a series of workloads and take memory metric measurements after
-	each launch.
+	Launch a series of workloads and take memory metric measurements using 'collectd'
+	after each launch.
```
grahamwhaley: In fact, we don't just take memory measurements any more - I think saying `a set of system level measurements using collected` will now be more accurate.

PR #279, opened 2019-11-28
scaling: add used memory to PDF report table

askervin:
Currently the PDF report shows how much scaling up consumes free
memory. This number is not comparable between cluster nodes or even
test runs because RAM used for OS caches/buffers/slab is counted as
consumed. As a consequence, consumed free memory depends heavily on
initial memory conditions of a node, instead of used memory by the k8s
and pods. This patch adds "memory used" to the report in order to have
less node-dependent and more reproducible memory figure.

Using /proc/meminfo "MemAvailable" was also tried out, but it varies
almost like "MemFree" that is currently reported.


grahamwhaley:
Looking good @askervin . Code looks good, and it ran and made a table for me... a couple of questions:
- should we also plot the used data on the graph? I suspect so, as having both used and and free in the table but not in the graph does not feel correct.
- we probably need to update the test text in the top level `.Rmd` file as well to match the changes (to say we are measuring not just free memory any more).

I also suspect once we have settled on which result is 'best' (most representative of theoretical maximum pod density), we should also note that in the `.Rmd` file text to make it clear which number to grab from the report...

askervin:
Good points, thanks @grahamwhaley. I added the memory used plot. Now the page looks more complete, and the title offered a nice place to explain what "used" actually means. (That is, why $\delta$Free > $\delta$Used.)

A note on grabbing used memory stats added to Rmd as well.

askervin:
rebased, commits squashed, tested, ready for merge.

grahamwhaley:
@marcemq - [PTAL](https://en.wiktionary.org/wiki/PTAL)

PR #280, opened 2019-11-29
metrics: improve view for high pod count data

grahamwhaley:
A collection of improvements for viewing larger and noisy data sets:
- If we have a lot of data points (>20), do not plot the geom_points, as they just add noise
- Use a linear model fitting to try and calculate the CPU-per-pod data
- Try to fit a colour palette to the test sets to separate them, if we can

### DIFF
--- a/metrics/report/report_dockerfile/collectd_scaling.R
+++ b/metrics/report/report_dockerfile/collectd_scaling.R
@@ -317,7 +317,27 @@ for (currentdir in resultdirs) {
 				}
 				min_idle_cpu=node_cpu_idle_data$value[cpu_end_index]
 
-				cputotal = cputotal + (max_idle_cpu - min_idle_cpu)
+				# Use a linear regression model to try and guesstimate the CPU
+				# cost per pod.
+				# We used to use the formula:
+				#  cputotal = cputotal + (max_idle_cpu - min_idle_cpu)
+				# to examine the difference from the first and last sample, but, the data
+				# for cpu is quite noisy. This could easily lead to the first/last samples
+				# being sat in a peak or trough, and thus throwing out the actual value.
+				# Using the linear regression, at least if our measurements are fairly linear
+				# then maybe we get a more realistic result.
+
+				cpu_lm=lm(value ~ epoch, data=node_cpu_idle_data[cpu_start_index:cpu_end_index,])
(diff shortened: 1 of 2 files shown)


grahamwhaley:
for the colour changes for instance, we are going from:

oldstyle

to:

newstyle

We can debate if we think that is an improvement... I don't actually like the palette as much (that is, the old pastel ones are quite 'nice'), but he separation and ability to visually match up related data is much better I think.

/cc @askervin @dklyle @obedmr @marcemq

metrics/report/report_dockerfile/collectd_scaling.R
```diff
@@ -314,7 +318,27 @@ for (currentdir in resultdirs) {
 				}
 				min_idle_cpu=node_cpu_idle_data$value[cpu_end_index]
 
-				cputotal = cputotal + (max_idle_cpu - min_idle_cpu)
+				# Use a linear regression model to try and guesstimate the CPU
```
grahamwhaley: I was going to note - if there is one bit of code that needs somebody else to really stare at it - it is this `lm()` usage - anybody out there got some good background in statistics?...

askervin: Wow, this is interesting... :)
I'll need to try with few test rounds to see real values that linear regression gives from the data I have.
In manual debugging I've used averages over few measurements at the start and end of the collectd cpu data, so this one is pretty advanced. :)

askervin:
This works fine for me. I'll still test `lm()` in more detail...

metrics/report/report_dockerfile/collectd_scaling.R
grahamwhaley: :-) Yeah, I wanted to work out how to 'fit a curve/line to the data I could then take measures from'. I'm definitely open to other suggestions as well.

grahamwhaley: A note in passing - there may be another way to extract the interpolated/intersection points using the R [predict](https://www.rdocumentation.org/packages/raster/versions/3.0-7/topics/predict) function. The following article is worth a read:

https://aosmith.rbind.io/2018/11/16/plot-fitted-lines/

grahamwhaley:
closing, work on hold.

PR #281, opened 2019-11-29
metrics: change collectd output to host /opt/collectd/run

askervin:
Currently we loose collectd data from a node when scaling ends to a
system failure on the node - yet this data can be very helpful in root
causing the failure. This patch changes collectd configuration so that
the output will be continuously written to host filesystem instead of
the collectd container overlay that will be lost unless scaling
reaches graceful exit.

### DIFF
--- a/metrics/collectd/collectd.bash
+++ b/metrics/collectd/collectd.bash
@@ -12,15 +12,25 @@ collectd_pod="collectd"
 init_stats() {
 	local wait_time=$1
 
-	# create collectd-config configmap
+	# create collectd-config configmap, delete old if there is one
+	kubectl get configmap collectd-config >/dev/null 2>&1 && kubectl delete configmap collectd-config
 	kubectl create configmap collectd-config --from-file=${COLLECTD_DIR}/collectd.conf
+
+	# if there is collectd daemonset already running, delete it
+	# to make sure that the latest configmap will be used.
+	kubectl get daemonset collectd >/dev/null 2>&1 && kubectl delete daemonset --wait=true --timeout=${delete_wait_time}s "${collectd_pod}"
 
 	# Launch our stats gathering pod
 	kubectl apply -f ${COLLECTD_DIR}/${collectd_pod}.yaml
 	kubectl rollout status --timeout=${wait_time}s daemonset/${collectd_pod}
 
+	# clear existing collectd output
+	while read -u 3 name node; do
+		kubectl exec -ti $name -- sh -c "rm -rf /mnt/opt/collectd/run/localhost/*"
+	done 3< <(kubectl get pods --selector name=collectd-pods -o json | jq -r '.items[] | "\(.metadata.name) \(.spec.nodeName)"')
+
 	# attempting to provide buffer for collectd to be installed and running,
-    # and CPU collection to build adequate history
+	# and CPU collection to build adequate history
 	sleep 12
 }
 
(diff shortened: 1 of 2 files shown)


metrics/collectd/collectd.bash
```diff
@@ -12,15 +12,21 @@ collectd_pod="collectd"
 init_stats() {
 	local wait_time=$1
 
-	# create collectd-config configmap
+	# create collectd-config configmap, delete old if there is one
+	kubectl get configmap collectd-config >/dev/null 2>&1 && kubectl delete configmap collectd-config
```
grahamwhaley: heh heh - yeah, sure, coz, how many times have we been caught out by this ;-)
I wonder if you want to do the same for the collectd daemonset as well?

metrics/collectd/collectd.conf
```diff
@@ -17,7 +17,7 @@ Hostname localhost
     ValuesPercentage true
 </Plugin>
 <Plugin "csv">
-    DataDir "/opt/collectd"
+    DataDir "/mnt/opt/collectd/run"
```
grahamwhaley: I don't see how `/mnt` is different from `/opt` - what am I missing? I don't see it put up as a volume in the pod for instance - probably some defacto setup or default I don't know about?

askervin: The collectd container has `host:/opt` mounted to `container:/mnt/opt` ([collectd.yaml](https://github.com/clearlinux/cloud-native-setup/blob/a0ca2a2017cb36624d3e34b846161a243d73a56c/metrics/collectd/collectd.yaml#L44)). `container:/opt`, where collectd has written the results so far, is located on overlay.

metrics/collectd/collectd.bash
askervin: Actually I'd prefer to leave the collectd daemonset creation as it is.

If the daemonset is already running and the configuration (collectd.yaml) has not changed, current code works fine:

```
# init_stats 10
configmap "collectd-config" deleted
configmap/collectd-config created
daemonset.apps/collectd unchanged
daemon set "collectd" successfully rolled out
```

On the other hand, if there's a change in yaml, for instance
```
# grep args -A1 *.yaml
          args:
            - echo my name is $(whoami) ; collectd -f;
```

current code nicely updates the existing daemonset
```
# init_stats 10
configmap "collectd-config" deleted
configmap/collectd-config created
daemonset.apps/collectd configured
Waiting for daemon set "collectd" rollout to finish: 1 out of 2 new pods have been updated...
Waiting for daemon set "collectd" rollout to finish: 1 out of 2 new pods have been updated...
Waiting for daemon set "collectd" rollout to finish: 1 of 2 updated pods are available...
daemon set "collectd" successfully rolled out
```

...and it indeed gets updated
```
# kubectl logs collectd-9qqsr
my name is root
plugin_load: plugin "aggregation" successfully loaded.
...
```

grahamwhaley: ah, OK, understood - only the configmap clashes on re-creation, but the daemonset updates (or not) as necessary. understood.

askervin: There's something still that I didn't really consider already. What happens if configmap has changed but yaml has not. That is, we'd like to change collectd configuration, but collectd daemonset happened to be already running... I'll test if we indeed should restart the daemonset anyway, just to make sure it's running the latest configuration.

askervin: Checked it. Collectd daemonset needed a restart in order to apply new configuration. I added daemonset restart to the commit and force-pushed new version.

So it was very good point after all, thanks! :)

grahamwhaley:
for reference, as @dklyle mentioned it in passing, just tested this on a 'kind' cluster (as it has significantly different structure to the other clusters we tend to test on).... and it worked fine.

grahamwhaley:
afaict, this was reviewed, acked, and tested... I'm going to merge it - and we'll have to fix it up later if we bust something.... :crossed_fingers:

PR #288, opened 2019-12-05
Make and rename net server process variables configurable

marcemq:
Some CNIs takes longer for its related deployments to become ready, that
is why `proc_wait_time` needs to be customized. Now `proc_wait_time` can
be set at execution time and has a default value too for time to pod network test harness.


askervin:
lgtm, too.

Issue #290, opened 2019-12-06
metrics: visualising launch time trend


grahamwhaley:
We had a request of the form:
> Q. If I wanted to answer the question 'how much extra overhead does pod launch 1001 have over pod launch 1000, how would I view that?'

I don't think we can answer (sensibly) that question directly, as the launch time data tends to be a bit noisy, so looking just at the launch between two pods in a test that large may not give an accurate representation.

I've put together a chart and table that split the data into sub-series of launches (by default, into 10 slices), calculates the 'mean diff' of all launches in that slice (that is, it gets the diff between each subsequent launch in the slice, and then gets the mean of those diffs), and plots those results as bars, overlayed with a smoothed line of the diff data itself.

This does give a better visual 'clue' to the launch time trends than just staring at the plain boot chart I think, *but*, I'm not yet 100% convinced this is the ultimate view of the data.

PR coming...

grahamwhaley:
PR was closed - closing the Issue.

PR #291, opened 2019-12-06
metrics: boot: add 'launchtime diff' page to report

grahamwhaley:
In order to help evaluate how the launch times of pods vary
over time, add a page that evaluates the mean diff of boot times
within slices of the whole sequence. By default, show the whole
boot sequence sliced into 10 evaluations.

We also overlay the smoothed line plot of the boot time diffs
themselves.

The code tries to accomodate test runs that may not divide nicely
into 10 buckets.

### DIFF
--- a/metrics/report/report_dockerfile/collectd_scaling.R
+++ b/metrics/report/report_dockerfile/collectd_scaling.R
@@ -18,6 +18,7 @@ testnames=c(
 )
 
 podbootdata=c()		# Track per-launch data
+bootmeandata=c()	# Calculate mean data across 'buckets'
 cpuidledata=c()		# Track cpu idle data per nodes
 memfreedata=c()		# Track mem free data for nodes
 inodefreedata=c()	# Track inode free data for nodes
@@ -28,7 +29,10 @@ iferrordata=c()		# Track interface errors data for nodes
 memstats=c()		# Statistics for memory usage
 cpustats=c()		# Statistics for cpu usage
 bootstats=c()		# Statistics for boot (launch) times
+bootmeanstats=c()	# Statistics for boot (launch) times mean bucket diffs
 inodestats=c()		# Statistics for inode usage
+
+bootdiffbuckets=10	# How many buckets do we evaluate the boot times in
 
 # iterate over every set of results (test run)
 for (currentdir in resultdirs) {
@@ -94,6 +98,41 @@ for (currentdir in resultdirs) {
 			# get the epoch time in seconds for the boot
 			local_bootdata$epoch = local_bootdata$ns/1000000000
 			local_bootdata$s_offset = local_bootdata$epoch - local_bootdata[1,]$epoch
+
+			# Grab the stepping used during testing, so we can calculate bucket sizes
+			stepping=fdata$Config$STEP
+
+			# Here we try to calculate and get a feel for 'changes over time' for
+			# launch times. That is, we are looking to see if the 'launch overheads
+			# per pod' change over time (or is it a linear cost?).
+			# We do this by taking the launch time diff between each consecutive launch,
+			# and then looking at the mean of those diffs across a 'buckets' width. We do
+			# this to de-noise the data (as actual individual diffs bounce up and down around
+			# the zero line a lot... so we are looking for the 'trend')
+
+			# Set up a diff of the consecutive boot times. Prepend with '0' so the result
+			# is the same length as the original data...
+			local_bootdata$launch_diff=c(0, diff(local_bootdata$launch_time))
+
+			# Annotate entries into 'buckets', so we can get some 'rolling' means
+			# Calculate the bucket mid point values of the range they cover, so they
+			# plot sanely along with n_pods based data.
+			bucket_size=ceiling(nrow(local_bootdata)/bootdiffbuckets)
+
+			# Take into account any STEP used in the testing.
+			pods_per_bucket=bucket_size*stepping
+			local_bootdata$bucket_min=(rep(0:(bootdiffbuckets-1), each=bucket_size)*bucket_size)[1:nrow(local_bootdata)] * stepping
+			local_bootdata$bucket_max=local_bootdata$bucket_min + (pods_per_bucket-1)
+			local_bootdata$bucket_mid=local_bootdata$bucket_min + (pods_per_bucket/2)
+
+			# And calculate the means diffs for those buckets
+			local_bootmean=aggregate(launch_diff ~ bucket_min, data=local_bootdata, FUN=mean)
+			local_bootmean$testname=rep(testname, nrow(local_bootmean))
+
+			# And annotate the bucket results with max/mid points as well, so we can
+			# form a stats table
+			local_bootmean$bucket_max=local_bootmean$bucket_min + (pods_per_bucket-1)
+			local_bootmean$bucket_mid=local_bootmean$bucket_min + (pods_per_bucket/2)
 
 			# Now Calculate some stats. This gets more complicated as we may have n-nodes,
 			# and we want to show a 'pod average', so we try to assess for all nodes. If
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Current output looks like below. This is with three data sets - of 100 or 50 pod runs, with steps of 1 or 10 (the test dataset names should be easy to interpret).

boot_diff

As noted in the Issue #290 , I'm not yet 100% convinced we could not present this data better (such as possibly looking at the diff *between* subsequent buckets).
Also, you'll see in this picture that anything with a STEP larger than 1 will bloat the graph vertical resolution, as the launch data is for 'n' launches to complete, not 1. That then squashes any smaller data and makes it less visible. Not a huge amount we can do there (well, OK, we could show as % change, which might help)...

In theory this code should work with pretty much any data set - it should trim/pad out any data entries where the bucket divisions do not fall nicely, and it should cope with STEP!=1.

/cc @askervin @marcemq @msivosuo

askervin:
@grahamwhaley , what would you think about having linear regression line(s) (`lm()`) in the figure to emphasize if there's a trend of some sort?

I like the idea of presenting start times with a fixed number of intervals (buckets) instead of depending on STEP size for the test.

Currently more explanation is needed to make clear the difference what is shown with lines and what with bars. Now it looks a bit confusing because a line can be clearly below zero, yet the bar is above.

image

grahamwhaley:
:smile: - really glad you are questioning @askervin - as I am/was not convinced we have the best view of this data yet... and first let me say - yes, it has been bugging me that the 'blocks' do not obviously align with the smoothed line - in fact, I've seen some that I'm not even sure I could match up in my visual head....

To further, and clarify hopefully, the discussion - I'm going to use a single 1000 pod step 10 dataset I happen to have around, which is hopefully less noisy than the other set(s) I was using...

Here is the basic boot time plot of that data:

boot_time

and the corresponding 'boot diff step' plot:

boot_steps

and then let's try a few things.... you have suggested `lm`, and I had already been wondering what a 'rolling mean` would look like (that is, the mean of the previous ten entries say, plotted).

I hacked up:

```R
library(zoo)
podbootdata$rmean = rollmean(podbootdata$launch_diff, 10, fill=c(0,0,0))

x <- ggplot(podbootdata) +
        geom_line(aes(n_pods, launch_diff, colour="diff"), alpha=0.5) +
        geom_line(aes(n_pods, rmean, colour="rmean")) +
        geom_line(aes(n_pods, launch_diff, colour="loess"),
                stat="smooth", se=FALSE, method="loess", span=0.25) +
        geom_line(aes(n_pods, launch_diff, colour="lm"),
                stat="smooth", se=FALSE, method="lm", span=0.25) +
        xlab("pods") +
        ylab("Mean launch time change (ms)") +
        ggtitle("Mean pod launch time change") +
        theme(legend.position="bottom") +
        theme(axis.text.x=element_text(angle=90))
```

which gets us:

rmean_and_lm


Let's discuss what that looks like, and how that might help or not with the original question of 'how can I tell what the cost difference is between the 1000 and 1001'th boot?' ...

Looking at the boot time chart, to me it seems we can see around the ~300 pod mark, there is a visible 'bump' in boot times. To me I'd expect that to then show up in the 'diff' chart. Looking at the combined chart above, the only line that is really hinting at that to me is the 'loess' line.

I threw the `lm` line in - the problem with (simple) `lm` is you are going to get a flat line, and that is probably not going to fit our data - well, that is not strictly true.... for pretty flat data, the `lm` will fit well, but then the blocks and the 'cost per extra pod` numbers in the table will all be pretty much the same anyway. In a noisy case, the `lm` is not going to fit, so I suspect will be more misleading than anything :-(

I'm sort of coming to the conclusion that the boot time data is too 'noisy' to actually usefully answer the posed '1000 to 1001th' question. tbh, I'd rather publish no numbers rather than misleading ones... I have considered rather than the 'blocks', we plot the `loess` line, and take 10 intercepts from it, and use those (as well as marking them on the line with tick marks)....

Let me see if I find time to mock that up.

metrics/report/report_dockerfile/collectd_scaling.R
```diff
@@ -28,8 +29,11 @@ iferrordata=c()		# Track interface errors data for nodes
 memstats=c()		# Statistics for memory usage
 cpustats=c()		# Statistics for cpu usage
 bootstats=c()		# Statistics for boot (launch) times
+bootmeanstats=c()	# Statistics for boot (launch) times mean bucket diffs
 inodestats=c()		# Statistics for inode usage
 
+bootdiffbuckets=10	# How many buckets do we evaluate the boot times in
```
marcemq: Depending on the amount of data, having the buckets fixed may hide a data distribution that can be useful to spot. I'm not sure if R have an automatic bucket recommendation, I'm sure that `matlab` and and `octave` does.

metrics/report/report_dockerfile/collectd_scaling.R
```diff
@@ -95,6 +99,41 @@ for (currentdir in resultdirs) {
...
+
+			# Here we try to calculate and get a feel for 'changes over time' for
+			# launch times. That is, we are looking to see if the 'launch overheads
+			# per pod' change over time (or is it a linear cost?).
+			# We do this by taking the launch time diff between each consequtive launch,
+			# and then looking at the mean of those diffs across a 'buckets' width. We do
+			# this to de-noise the data (as actual individual diffs bounce up and down around
+			# the zero line a lot... so we are looking for the 'trend')
+
+			# Set up a diff of the consequtive boot times. Prepend with '0' so the result
```
marcemq: Typo at `consequtive`

metrics/report/report_dockerfile/collectd_scaling.R
```diff
@@ -95,6 +99,41 @@ for (currentdir in resultdirs) {
...
+			# Annotate entries into 'buckets', so we can get some 'rolling' means
+			# Calculate the bucket mid point values of the range they cover, so they
+			# plot sanely along with n_pods based data.
+			bucket_size=ceiling(nrow(local_bootdata)/bootdiffbuckets)
+
+			# Take into account any STEP used in the testing.
+			pods_per_bucket=bucket_size*stepping
+			local_bootdata$bucket_min=(rep(0:(bootdiffbuckets-1), each=bucket_size)*bucket_size)[1:nrow(local_bootdata)] * stepping
+			local_bootdata$bucket_max=local_bootdata$bucket_min+(pods_per_bucket-1)
+			local_bootdata$bucket_mid=local_bootdata$bucket_min + (pods_per_bucket/2)
```
marcemq: nitpicky: user (or not) space between main addition operation.

marcemq:
I think we need to define a common ground to showcase this `launch time diff` first, and mostly when having more than one report since each can have its own used `STEP`, so I propose to merge results when the used `STEP` is the same (I think the number of pods will not affect the graph to be shown, it will only make it larger towards axis `x`).

metrics/report/report_dockerfile/collectd_scaling.R
grahamwhaley: ack. fixed, in two places.

metrics/report/report_dockerfile/collectd_scaling.R
grahamwhaley: ack. good spot. went with spaces for this patch.

metrics/report/report_dockerfile/collectd_scaling.R
grahamwhaley: Absolutely. It's also an issue if an 'interesting data point' happens to fall on a boundary between two slices. It is a very similar question to how much 'smoothing' we do to the data with the smoothed or rolling mean lines. Note, this bucketing (possibly better called slicing?) code is **not** a histogram - we are bucketing the linear (time based) data.
There is also the issue of if we get more buckets, they won't fit into the text table. The general idea of the page is to give a 'quick overview' to the questions such as:
- what is the cost per-container between launch `n` and `n+1`?
- is the launch time overhead fairly linear and constant?

If we take bucket number selection to the extremes then we have:
- at minimum, we would show one bucket. That is basically the same as using a single variable `lm()` (flat line), or the mean of the whole data set. That's fine if the data is linear, but I don't think our data is generally that linear
- at maximum, we would show one bucket per sample. That is the equivalent of plotting out the raw 'diff' data - see the graph I posted above - that data is pretty noisy.

I'm probably coming to the conclusion that buckets are not the ideal solution, and that some form of smoothing or rolling mean is better - visually it gives a much better representation of the 'average' boot diff time for each sample point, and not just for 10 individual sub-ranges. I'm leaning towards not calculating/drawing the buckets - but using a smoothed line, and taking 10 sample points off that line for the table data [iyswim](https://acronyms.thefreedictionary.com/IYSWIM).

OOI, I did have a look around for what R might offer in the 'auto bucketing' methods. Yes, there are some, and they tend to be used with the [histograms](https://stat.ethz.ch/R-manual/R-devel/library/graphics/html/hist.html). See the [nclass()](https://stat.ethz.ch/R-manual/R-devel/library/grDevices/html/nclass.html) functions for instance.
If we were going to auto-bucket then I think we'd use those. I may have a play with them, but I have a feeling at some point they would return a large number, and that would blow the text table width out of the page...

grahamwhaley:
Closing. Unclear if this view really helped, and work is now on hold.

PR #292, opened 2019-12-11
Add the minimun connection time out for `curl` call

marcemq:
The time to pod network test harness was hanging when the net server was
not ready due to the `curl` execution because `curl` has its own connection
time out and it can be up to two minutes. Now, such connection time out is set
to one second.

### DIFF
--- a/metrics/scaling/k8s_scale_net.sh
+++ b/metrics/scaling/k8s_scale_net.sh
@@ -165,7 +165,7 @@ run() {
 		info "IP: $IP"
 
 		# service health check
-        cmd="curl --noproxy \"*\" http://$IP:8080/healthz"
+        cmd="curl --noproxy \"*\" http://$IP:8080/healthz --connect-timeout 1"
         waitForProcess "$proc_wait_time" "$proc_sleep_time" "$cmd" "http server is not ready yet!!"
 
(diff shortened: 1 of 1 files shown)


askervin:
nc measurements, which do pretty much the same without http, have recorded > 1 s timeouts when running > 400 pods with Canal CNI. After 700 pods, already more than 5 % of round trip times (including tcp/ip connect-send-recv-close like here as well) have exceeded 1 s.

In case timeout leads to a test failure, I'd vote for a longer timeout. On the other hand, if it will not fail the test, we'll just get faster test execution and prettier graph than in nc tests because of not having to scale y axis to 8000 ms, for instance. :)

askervin:
@marcemq, I tested this by causing timeouts to the curl request by rebooting the node where net-serve pod is running and by suspending the agnhost process (`kill -STOP $(pidof agnhost)`).

I'd suggest adding `--max-time 1` after `--connect-timeout 1`, because otherwise curl was still able to hang.

I'd also suggest adding net-serve* service cleanups in the `cleanup()` routine. Currently services can be left behind after failed retries due to `curl` timeouts, because `set -e` forces exit when `waitForProcess` fails. Services that are left behind will fail the next test run.

PR #294, opened 2019-12-13
metrics: add memory load generator

askervin:

### DIFF
--- a/metrics/lib/mem-load.bash
+++ b/metrics/lib/mem-load.bash
@@ -1,0 +1,137 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Helper routines for setting up a constant MEM load on the cluster/nodes
+
+MEMLOAD_DIR=${THIS_FILE%/*}
+
+# Default load: 50% of free memory per worker, run 1 worker
+SMF_MEM_LOAD_WORKER_BYTES=${SMF_MEM_LOAD_WORKER_BYTES:-80%}
+SMF_MEM_LOAD_MASTER_BYTES=${SMF_MEM_LOAD_MASTER_BYTES:-80%}
+SMF_MEM_LOAD_WORKER_VMS=${SMF_MEM_LOAD_WORKER_VMS:-1}
+SMF_MEM_LOAD_MASTER_VMS=${SMF_MEM_LOAD_MASTER_VMS:-1}
+# Default memory limits or requests: none
+SMF_MEM_LOAD_WORKER_LIMIT=${SMF_MEM_LOAD_WORKER_LIMIT:-}
+SMF_MEM_LOAD_MASTER_LIMIT=${SMF_MEM_LOAD_MASTER_LIMIT:-}
+SMF_MEM_LOAD_WORKER_REQUEST=${SMF_MEM_LOAD_WORKER_REQUEST:-}
+SMF_MEM_LOAD_MASTER_REQUEST=${SMF_MEM_LOAD_MASTER_REQUEST:-}
+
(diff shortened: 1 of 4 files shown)


askervin:
@grahamwhaley , @marcemq , @dklyle,

According to my first tests, memory load generator helps us to drive cluster very fast to similar state where we can seeing long delays just like with high pod counts.

I'd like to add here extra configuration option to skip loading master node. Do you think it would be ok?

grahamwhaley:
adding config options to skip loading master nodes sounds perfect to me. In fact, we should probably make both master and worker nodes configurable independently - I can see we may want to load either or both for testing.

askervin:
@grahamwhaley, separate configuration for worker and master sounds very good. It's implemented in the latest commit. Tested in two and three-node clusters, but not in single-node cluster.

metrics/lib/mem-load.bash
```diff
@@ -0,0 +1,137 @@
...
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Helper routines for setting up a constant MEM load on the cluster/nodes
+
+MEMLOAD_DIR=${THIS_FILE%/*}
+
+# Default load: 50% of free memory per worker, run 1 worker
```
grahamwhaley: s/50/80/

metrics/lib/mem-load.md
```diff
@@ -0,0 +1,52 @@
...
+```bash
+export SMF_MEM_LOAD_MASTER=true
+export SMF_MEM_LOAD_WORKER=true
+```
+
+### 64 GB load on worker node(s)
+
+This configuration reserves 64 GB of memory on every worker node. The
+stress-ng pods request 64 GB memory resource. Deploying the DaemonSet
+will not finish unless all available workers match the request.
```
grahamwhaley: ... 'within the `wait_time` timeout period' ?

metrics/lib/mem-load.md
```diff
@@ -0,0 +1,52 @@
+# `mem-load` stack stresser
```
grahamwhaley: We should ref this doc from the top level doc, like we do for the cpu-load: https://github.com/clearlinux/cloud-native-setup/tree/master/metrics#cpu-load-generator

metrics/lib/mem-load.bash
```diff
@@ -0,0 +1,137 @@
...
+
+	# And store off our config into the JSON results
+	metrics_json_start_array
+	local json="$(cat << EOF
+	{
+		"LOAD_WORKER": "${SMF_MEM_LOAD_WORKER}",
+		"LOAD_WORKER_BYTES": "${SMF_MEM_LOAD_WORKER_BYTES}",
+		"LOAD_WORKER_VMS": "${SMF_MEM_LOAD_WORKER_VMS}",
+		"LOAD_WORKER_LIMIT": "${SMF_MEM_LOAD_WORKER_LIMIT}",
+		"LOAD_WORKER_REQUEST": "${SMF_MEM_LOAD_WORKER_REQUEST}"
```
grahamwhaley: `,` missing on the end of `LOAD_WORKER_REQUEST` line, which makes invalid JSON.
That gives the following error in the report:

```
## Error in parse_con(txt, bigint_as_char): parse error: after key and value, inside map, I expect ’,’ or ’}’
##
"LOAD_WORKER_REQUEST": ""
"LOAD_MASTER": "",
"LOAD_MASTE
##
(right here) ------^
```

grahamwhaley:
Hi @askervin ! - given this is no longer active, could you close this PR down for us? thx!

askervin:
Ah, sure, thanks @grahamwhaley!

Issue #297, opened 2020-01-09
setup_system.sh fails when OS version is specified and packages have been removed


CraigSterrett:
Running setup_system.sh on a system setting the OS version to keep setup_system from upgrading the OS causes an error because a package has been removed. Need to add the --force flag to the 
sudo swupd repair -m "${CLR_VER}" --picky command

Error received is below
Setting OS Version...
Diagnosing version 31890
Warning: The --picky option is specified; ignoring version mismatch for repair
Downloading missing manifests...
Warning: Bundle "sshfs" is invalid, skipping it...

Error: Unable to verify. One or more currently installed bundles are not available at version 31890. Use --force to override

Repair did not fully succeed

PR #298, opened 2020-01-09
Added --force flag to swupd repair command

CraigSterrett:
Running setup_system.sh on a system setting the OS version to keep setup_system from upgrading the OS causes an error because a package has been removed. Need to add the --force flag to the
sudo swupd repair -m "${CLR_VER}" --picky command

Closes issue #297


ahsan518:
Testing

Issue #306, opened 2020-02-05
rook: fails to come up on single node cluster


grahamwhaley:
Setting up a single node cluster (that is, not joining any slave nodes at the point where the setup pauses for you to) of 'all' using create_stack.sh, the process gets stuck 'forever' trying to bring up rook:

```
...
rolebinding.rbac.authorization.k8s.io/rook-ceph-mgr created
rolebinding.rbac.authorization.k8s.io/rook-ceph-osd created
rolebinding.rbac.authorization.k8s.io/rook-ceph-system created
clusterrolebinding.rbac.authorization.k8s.io/cephfs-csi-nodeplugin created
clusterrolebinding.rbac.authorization.k8s.io/cephfs-csi-provisioner-role created
clusterrolebinding.rbac.authorization.k8s.io/rbd-csi-nodeplugin created
clusterrolebinding.rbac.authorization.k8s.io/rbd-csi-provisioner-role created
clusterrolebinding.rbac.authorization.k8s.io/rook-ceph-system-psp created
clusterrolebinding.rbac.authorization.k8s.io/rook-csi-cephfs-plugin-sa-psp created
clusterrolebinding.rbac.authorization.k8s.io/rook-csi-cephfs-provisioner-sa-psp created
clusterrolebinding.rbac.authorization.k8s.io/rook-csi-rbd-plugin-sa-psp created
clusterrolebinding.rbac.authorization.k8s.io/rook-csi-rbd-provisioner-sa-psp created
clusterrolebinding.rbac.authorization.k8s.io/rook-ceph-global created
clusterrolebinding.rbac.authorization.k8s.io/rook-ceph-mgr-cluster created
clusterrolebinding.rbac.authorization.k8s.io/rook-ceph-object-bucket created
clusterrolebinding.rbac.authorization.k8s.io/rook-ceph-osd created
deployment.apps/rook-ceph-operator created
cephblockpool.ceph.rook.io/replicapool created
cephcluster.ceph.rook.io/rook-ceph created
podsecuritypolicy.policy/rook-privileged created
No resources found
Waiting for Rook OSD
Waiting for Rook OSD
^C
```

For ref:

```
$ sudo swupd update -s
Current OS version: 32230
Latest server version: 32230
There are no updates available
```

I don't know if that is specific to my system, or due to being a single node. I'm pretty sure this used to work. If you need me to re-try etc., then tell me what to try and I'll log back here.

CraigSterrett:
[...]

PR #310, opened 2020-03-23
rook: allow multiple mons per node

ganeshmaharaj:
On a single node setup, rook will fail to get stuck waiting on OSDs as
monitors by default have an anti-affinity rule that prevents multiple
monitors to land on a single node. This will also be an issue if you are
using this setup of < 4 nodes (1 master and 3 workers) as the master has
the NoSchedule taint. This patch allows multiple monitors on a single
node. When multiple nodes are present, the monitor pods are
automatically distributed.

Downside: When the setup is started with a single node and then more
nodes are added, the monitor pods *DO NOT* automatically re-balance.

Fixes: #306

### DIFF
--- a/clr-k8s-examples/7-rook/overlays/v1.1.0/patch_cephcluster.yaml
+++ b/clr-k8s-examples/7-rook/overlays/v1.1.0/patch_cephcluster.yaml
@@ -4,6 +4,17 @@ metadata:
   name: rook-ceph
   namespace: rook-ceph
 spec:
+  mon:
+    allowMultiplePerNode: true
   storage:
     directories:
      - path: /var/lib/rook
+---
+apiVersion: ceph.rook.io/v1
+kind: CephBlockPool
+metadata:
+  name: replicapool
+  namespace: rook-ceph
+spec:
+  replicated:
+    size: 1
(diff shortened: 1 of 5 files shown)


clr-k8s-examples/7-rook/overlays/v1.1.0/patch_cephcluster.yaml
```diff
@@ -4,6 +4,8 @@ metadata:
   name: rook-ceph
   namespace: rook-ceph
 spec:
+  mon:
+    allowMultiplePerNode: true
   storage:
     directories:
-     - path: /var/lib/rook
\ No newline at end of file
+     - path: /var/lib/rook
```
grahamwhaley: hehe, spurious editor injected extra line feeds? ;-)

ganeshmaharaj: Bah! fixing it. :)

ganeshmaharaj:
@CraigSterrett @grahamwhaley  all comments addressed. Let me know if this looks good now.

CraigSterrett:
Replacing with a PR which adds separate configurations for single node installs and multinode installs

Issue #311, opened 2020-03-24
Add information about Rook to readme


CraigSterrett:
The readme needs information about Rook and should contain a disclaimer that this is an example setup not to be used in production

PR #313, opened 2020-03-24
Fix rook single node setup

CraigSterrett:
Modified the rook installation to support both multinode kubernetes
clusters and standalone kubernetes clusters. Multinode installations
will occur as before, with changes for standalone installations.
v0.8.3 I found it was not currently working and
is too old to continue to support, will submit a PR for removing it. Also made a couple minor spacing changes to yaml as detected by yamllint

closes issue 306 https://github.com/clearlinux/cloud-native-setup/issues/306
closes issue 311 https://github.com/clearlinux/cloud-native-setup/issues/311

Signed-off-by: Craig Sterrett <craig.Sterrett@intel.com>


grahamwhaley:
I think the code changes lgtm, but Ganesh is the master ;-)
I also think I'm missing a subtlety - how does the kustomize choose whether to apply the singlenode or multinode - I didn't spot where the decision gets made, so maybe it is an implicit kustomize thing? (I find the kustomize docs hard to find/read/lacking....)

CraigSterrett:
Dang, the reason you can't see how the kustomize is done is because my push missed that file, create_stack.sh has to be modified, the kustomize is called from within that script.

clr-k8s-examples/create_stack.sh
```diff
@@ -108,6 +109,7 @@ function cluster_init() {
 	#Ensure single node k8s works
 	if [ "$(kubectl get nodes | wc -l)" -eq 2 ]; then
 		kubectl taint nodes --all node-role.kubernetes.io/master-
+		mode="standalone"
```
ganeshmaharaj: Would this change work when a user does not use `create_stack.sh all` and creates the cluster one piece at a time? For eg: if i do `create_stack.sh init`, `create_stack.sh cni` and `create_stack.sh storage`, i might end up a hung cluster, correct? Would it help if we do the check of whether we have a single node or more right before we apply rook and then choose which set of manifests to apply?

PR #314, opened 2020-03-26
removed v0.8.3 of rook

CraigSterrett:
Rook v0.8.3 was found to be not working and is too old to continue
supporting.

Signed-off-by: Craig Sterrett <craig.Sterrett@intel.com>


NitinAtIntel:
lgtm

PR #320, opened 2020-04-28
metrics: Use a specific version of rocker/tidyverse

GabyCT:
This PR uses a specific version of rocker/tidyverse as the latest version
does not have the latex-xcolor package which makes impossible to create
the metrics report.

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>


grahamwhaley:
Oh, apart from, nominally we should also bump the Dockerfile version @GabyCT , so if anybody is using this then it will force a rebuild/use.

GabyCT:
@grahamwhaley done :)

PR #324, opened 2020-05-06
Switch private network for Vagrant VMs

ganeshmaharaj:
Previously used IP range seems to be allocated to NASA and probably not
a good idea to use it here. Switching the private network to be part of
the IPs allocated to private networks.


ganeshmaharaj:
@CraigSterrett can we land this change in? Anything blocking it?

PR #327, opened 2020-05-13
Make metrics pdf error reporting more graceful

grahamwhaley:
As the metrics report gained the ability to process more test results, it never gained the ability to cope with missing test data (for instance, if you only run a single test as that is the only one you are interested in). This led to confusion and mess, as the report would end up littered with cryptic truncated R error messages, which made folks think something had gone wrong with the test/data they were actually interested in.

Clean up the R code so it handles missing files and data more gracefully, and places meaningful 'nothing found' messages into the report.

To do this involved nesting the R code into functions, as there is no other way I could find to cleanly drop out of an Rmarkdown fragment without halting the processing of the rest of the pipeline, which is not what we needed.


grahamwhaley:
@amshinde @chavafg @gabyct - fyi.

@dklyle for old times sake and interest/amusement factor :-)

PR #333, opened 2020-12-17
Disable swap before OS Update

mdhorn:
Mask all swap targets.

This ensures no swap will be active to allow K8 testing


CraigSterrett (approved):
LGTM

PR #341, opened 2022-04-04
Remove static cpu from kubeadm.yaml

jascott1:
Removing static cpu pinning due to confusion and complications it causes with disabling/enabling cores.


ganeshmaharaj:
@jascott1  Mind adding this to the commit message?

```
Static CPU policy by defaults pins the process to CPUs if a kubernetes pod limits are set to be full cpus. While this a neat thing, this also causes issues when setaffinity is tried in a particular CPU. While we would like to set limits, pinning a process to a core is something we will try to avoid in the near future.
```

ganeshmaharaj:
:tada: Thanks for fixing this up.

ganeshmaharaj (approved):

PR #347, opened 2022-09-21
Fix create_stack for version 1.25

pwmorreale:
kubernetes 1.25 changed the key for the NoSchedule taint to 'control-plane'.

Fix the script to handle both pre and post version 1.25

Signed-off-by: Peter W. Morreale <pwmorreale@gmail.com>

### DIFF
--- a/clr-k8s-examples/create_stack.sh
+++ b/clr-k8s-examples/create_stack.sh
@@ -115,9 +115,17 @@ function cluster_init() {
 		fi
 	fi
 
-	#Ensure single node k8s works
+	#Ensure single node k8s works both pre and post v1.25
 	if [ "$(kubectl get nodes | wc -l)" -eq 2 ]; then
-		kubectl taint nodes --all node-role.kubernetes.io/master-
+		minor=$(kubeadm version -o short | cut -f 2 -d "." )
+		if [ $minor -ge "25" ]; then
+			kubectl taint nodes --all  node-role.kubernetes.io/control-plane-
+		elif [ $minor -eq "24" ]; then
+			kubectl taint nodes --all  node-role.kubernetes.io/control-plane-
+			kubectl taint nodes --all  node-role.kubernetes.io/master-
+		else
+			kubectl taint nodes --all  node-role.kubernetes.io/master-
+		fi
 		mode="standalone"
 	fi
 }


comjf:
Ha I should have checked the PRs before attempting my own fix here. I ran into the same issue, here is my solution that does the same thing:
```
diff --git a/clr-k8s-examples/create_stack.sh b/clr-k8s-examples/create_stack.sh
index e0a999a..60f12e2 100755
--- a/clr-k8s-examples/create_stack.sh
+++ b/clr-k8s-examples/create_stack.sh
@@ -117,7 +117,13 @@ function cluster_init() {
 
        #Ensure single node k8s works
-               kubectl taint nodes --all node-role.kubernetes.io/master-
+        if [[ -t 0 ]]; then
+            if [[ -n "${K8S_VER}" && $(echo "${K8S_VER}" | awk -F. '{print $2}') -lt 24 ]]; then
+                kubectl taint nodes --all node-role.kubernetes.io/master-
+            else
+                kubectl taint nodes --all node-role.kubernetes.io/control-plane-
+            fi
+        fi
                mode="standalone"
        fi
 }
```

I think yours is a little better. However, I think the version change happened in 1.24 no? 

I get that impression from reading: 
- https://github.com/kubernetes/kubeadm/issues/2200
- https://groups.google.com/g/kubernetes-sig-cluster-lifecycle/c/XJhn8bpvHac/m/ZP1sN_20BwAJ?pli=1

pwmorreale:
That might be correct. IIUC it was introduced in 1.24 and ‘master’ was removed in 1.25. 

Unsure whether that actual matters for this distro since it apparently jumped to 1.25???    Did CL ever release 1.24?  

Unfortunately I’m unable to check atm 

Best,
-PWM

dklyle (commented):
Seems there was a 1.24 release: https://github.com/clearlinux-pkgs/kubernetes/releases/tag/1.24.4-136. If you update to use 24 instead of 25, I'll approve.

pwmorreale:
AFAICT, 1.24 includes both taints for new clusters, 1.25 only contains the 'control-plane' taint, and everything prior to 1.24 contains the 'master' taint.  

https://github.com/kubernetes/kubernetes/pull/107533

dklyle (approved):
LGTM, thanks!

